PR #9, opened 2015-01-08
Added support for specifying cache directory on the command line.

muonzoo:
This allows for specifying the cache dir manually - which is helpful in test and multi-profile support. (Running two configs each with their own cache).


muonzoo:
Thanks!

Issue #12, opened 2015-03-29
Include the URL in the subject of the alert e-mail


thp:
From user feedback:

was wondering if there is a way to include the name of the URL changed in the subject of the alert email?

egroeper:
As URLs tend to be long perhaps this could be combined with shortnames for URLs/websites?
Then these shortnames could be put in the subject.
Of course this would mean to change the syntax of urls.txt.

thp:
Yeah, I would have no problem with changing urls.txt to JSON or YAML or something (and use the opportunity to add new settings and make it extendable) - this can work nicely with changing the filename and checking for the existence of both filenames (e.g. urls.txt and urls.yaml) and making sure that existing urls.txt files continue to work as before (just not with the advanced features), so users can upgrade to the new format at their leisure.

PR #24, opened 2015-10-05
Extendable urls file

egroeper:
My try to implement the extendable file format discussed in #12.

Please review.

### DIFF
--- a/lib/urlwatch/handler.py
+++ b/lib/urlwatch/handler.py
@@ -58,8 +58,9 @@ def get_current_user():
         return pwd.getpwuid(os.getuid()).pw_name
 
 class JobBase(object):
-    def __init__(self, location):
+    def __init__(self, location, name=None):
         self.location = location
+        self.name = name
 
     def __str__(self):
         return self.location
@@ -176,18 +177,18 @@ def shelljob_security_checks(urls_src):
     shelljob_errors = []
     current_uid = os.getuid()
 
-    dirname = os.path.dirname(urls_txt) or '.'
+    dirname = os.path.dirname(urls_src) or '.'
     dir_st = os.stat(dirname)
     if (dir_st.st_mode & (stat.S_IWGRP | stat.S_IWOTH)) != 0:
         shelljob_errors.append('%s is group/world-writable' % dirname)
     if dir_st.st_uid != current_uid:
         shelljob_errors.append('%s not owned by %s' % (dirname, get_current_user()))
 
-    file_st = os.stat(urls_txt)
+    file_st = os.stat(urls_src)
     if (file_st.st_mode & (stat.S_IWGRP | stat.S_IWOTH)) != 0:
-        shelljob_errors.append('%s is group/world-writable' % urls_txt)
+        shelljob_errors.append('%s is group/world-writable' % urls_src)
     if file_st.st_uid != current_uid:
-        shelljob_errors.append('%s not owned by %s' % (urls_txt, get_current_user()))
+        shelljob_errors.append('%s not owned by %s' % (urls_src, get_current_user()))
 
     return shelljob_errors
 
@@ -217,6 +218,44 @@ def parse_urls_txt(urls_txt):
                 sys.exit(1)
         else:
             jobs.append(UrlJob(line))
+
+    return jobs
+
+def parse_urls_yaml(urls_yaml):
+    import yaml
+    jobs = []
+
+    # Security checks for shell jobs - only execute if the current UID
+    # is the same as the file/directory owner and only owner can write
+    allow_shelljobs = True
+    shelljob_errors = shelljob_security_checks(urls_yaml)
+    if shelljob_errors:
+        allow_shelljobs = False
+
+    with open(urls_yaml, 'r') as f:
+        yaml_jobs = yaml.load_all(f)
+        for job in yaml_jobs:
+            # entries with all parameters commented out
+            if job is None:
+                continue
+            # set all supported keys here, so we don't need to care later
+            for key in ['name']:
+                if not key in job:
+                    job[key] = None
+            if not "url" in job:
+                print >>sys.stderr, '\n  Entry has no url! Stopping!\n'
+                sys.exit(1)
+            if job['url'].startswith('|'):
+                if allow_shelljobs:
+                    jobs.append(ShellJob(job['url'][1:], name=job['name']))
+                else:
+                    print >>sys.stderr, '\n  SECURITY WARNING - Cannot run shell jobs:\n'
+                    for error in shelljob_errors:
+                        print >>sys.stderr, '    ', error
+                    print >>sys.stderr, '\n  Please remove shell jobs or fix these problems.\n'
+                    sys.exit(1)
+            else:
+                jobs.append(UrlJob(job['url'], name=job['name']))
 
     return jobs
 
--- a/lib/urlwatch/handler.py
+++ b/lib/urlwatch/handler.py
@@ -276,22 +276,21 @@ def parse_urls_yaml(urls_yaml):
     return jobs
 
 def create_urls_yaml(jobs, urls_yaml):
+    yaml_jobs = []
     with os.fdopen(os.open(urls_yaml, os.O_WRONLY | os.O_CREAT, 0o0644), 'w') as f:
         for job in jobs:
             if isinstance(job, ShellJob):
-                f.write("---\n")
-                f.write("command: \"%s\"\n" % job.location)
+                yaml_jobs.append({"command": job.location})
             elif isinstance(job, UrlJob):
                 post_data = None
                 if ' ' in job.location:
                     location, post_data = job.location.split(' ', 1)
                 else:
                     location = job.location
-
-                f.write("---\n")
-                f.write("url: \"%s\"\n" % location)
+                yaml_job = {"url": location}
                 if post_data is not None:
-                    f.write("data: \"%s\"\n" % post_data)
+                    yaml_job["data"] = post_data
+                yaml_jobs.append(yaml_job)
             else:
                 log.warning("import: invalid job %s" % job)
-        f.write("---\n")
+        yaml.dump_all(yaml_jobs, stream=f, default_flow_style=False)
--- a/lib/urlwatch/handler.py
+++ b/lib/urlwatch/handler.py
@@ -276,3 +276,23 @@ def parse_urls_yaml(urls_yaml):
 
     return jobs
 
+def create_urls_yaml(jobs, urls_yaml):
+    with os.fdopen(os.open(urls_yaml, os.O_WRONLY | os.O_CREAT, 0o0644), 'w') as f:
+        for job in jobs:
+            if isinstance(job, ShellJob):
+                f.write("---\n")
+                f.write("command: \"%s\"\n" % job.location)
+            elif isinstance(job, UrlJob):
+                post_data = None
+                if ' ' in job.location:
+                    location, post_data = job.location.split(' ', 1)
+                else:
+                    location = job.location
+
+                f.write("---\n")
+                f.write("url: \"%s\"\n" % location)
+                if post_data is not None:
+                    f.write("data: \"%s\"\n" % post_data)
+            else:
+                log.warning("import: invalid job %s" % job)
+        f.write("---\n")
--- a/lib/urlwatch/handler.py
+++ b/lib/urlwatch/handler.py
@@ -258,12 +258,12 @@ def parse_urls_yaml(urls_yaml):
             for key in ['name', 'data']:
                 if not key in job:
                     job[key] = None
-            if not "url" in job:
-                print >>sys.stderr, '\n  Entry has no url! Stopping!\n'
-                sys.exit(1)
-            if job['url'].startswith('|'):
+
+            if "url" in job:
+                jobs.append(UrlJob(job['url'], name=job['name'], data=job['data']))
+            elif "command" in job:
                 if allow_shelljobs:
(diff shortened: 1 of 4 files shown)


egroeper:
Based on this I implemented a variant of #12 (showing titles/names of urls, not the urls themselves):
https://github.com/egroeper/urlwatch/tree/name_in_subject

But I think it only make sense to make pull request for that, if yaml support is done.
But perhaps it's nice for trying urls.yaml.

lib/urlwatch/handler.py
```diff
@@ -216,3 +221,41 @@ def parse_urls_txt(urls_txt):
 
     return jobs
 
+def parse_urls_yaml(urls_yaml):
+    import yaml
```
thp: Do the import once globally at the start of the module, not here.

share/urlwatch/examples/urls.yaml.example
```diff
@@ -0,0 +1,50 @@
...
+url: "|ls -al ~"
+---
+
+# If you want to use spaces in URLs, you have to URL-encode them (e.g. %20)
+url: "http://example.org/With%20Spaces/"
+---
+
+# You can do POST requests by writing the POST data behind the URL,
+# separated by a single space character. POST data is URL-encoded.
+url: "http://example.com/search.cgi button=Search&q=something&category=4"
```
thp: For POST data, shouldn't there be e.g. a `data:` field in the YAML?

urlwatch
```diff
@@ -109,11 +111,11 @@ class NullHandler(logging.Handler):
 
 log.addHandler(NullHandler())
 
-ERROR_MESSAGE_URLS_TXT = """
-Error: You need to create a urls.txt file first.
+ERROR_MESSAGE_URLS = """
+Error: You need to create either a urls.yaml or a urls.txt file first.
```
thp: What if both exist?

lib/urlwatch/handler.py
egroeper: No problem. Idea was to not introduce a new dependency, if the new method / config file type is not used.

share/urlwatch/examples/urls.yaml.example
egroeper: Good idea.

urlwatch
egroeper: It doesn't make sense to have both. Therefore I'm saying the user needs to have one of them, if he has none. The message is only shown, if there is no urls file.

yaml file as the newer format is given precedence, as stated in the code. But I don't think it's helpful to say that here. To keep it simple I'm saying the user he should use one and not what I'm doing, if he thinks having two is a good idea. Of course I'm open for suggestions.

Perhaps we should think about another place to document the precedence?

thp: Or we only have `urls.yaml` and we provide a way to migrate any old `urls.txt` to `urls.yaml` on first run (if `urls.yaml` doesn't exist yet) so that there's no confusion on which format to use.

egroeper: I'm fine with that. 
My suggestion:
Try to convert urls.txt to urls.yaml, if no urls.yaml exists. To be one the safe side keep urls.txt, but rename it to urls.txt.imported.

In the case there both (urls.txt and urls.yaml) give a warning suggesting the user should import manually. That way we avoid having to merge urls.txt into an existing urls.yaml.

egroeper:
Please review current state of this pr. I think I handled all your remarks.

lib/urlwatch/handler.py
```diff
@@ -37,8 +37,15 @@
...
 
+# Python 2 and 3 compatible string check
+try:
+    basestring
+except NameError:
+    basestring = str
+
 import subprocess
 import email.utils
+from urllib import urlencode
```
thp: Please use `import urllib` here and `urllib.urlencode` below to be consistent with the other imports here.

lib/urlwatch/handler.py
```diff
@@ -127,11 +140,16 @@ def retrieve(self, timestamp=None, filter_func=None, headers=None,
...
             log.info('Sending POST request to %s', self.location)
-        else:
-            post_data = None
+            # data might be dict or urlencoded string
+            if isinstance(self.data, dict):
+                # convert to urlencoded string
+                self.data = urlencode(self.data)
+            elif not isinstance(self.data, basestring):
+                # nuke / ignore other data (no string, no dict)
+                log.warning("invalid data parameter for url %s" % self.location)
```
thp: 1. Use proper capitalization
2. For log calls, let the logger handle string formatting (`,` instead of `%`)
3. Also tell the user what the invalid parameter is and that we are ignoring it

Use:

    log.warning('Ignoring invalid data parameter for URL %s: %r', self.location, self.data)

lib/urlwatch/handler.py
```diff
@@ -216,3 +238,61 @@ def parse_urls_txt(urls_txt):
...
+                if allow_shelljobs:
+                    jobs.append(ShellJob(job['command'], name=job['name']))
+                else:
+                    print >>sys.stderr, '\n  SECURITY WARNING - Cannot run shell jobs:\n'
+                    for error in shelljob_errors:
+                        print >>sys.stderr, '    ', error
+                    print >>sys.stderr, '\n  Please remove shell jobs or fix these problems.\n'
+                    sys.exit(1)
+            else:
+                print >>sys.stderr, '\n  Entry has neither url nor command! Stopping!\n'
```
thp: Instead of stopping, maybe print a warning instead and continue?

lib/urlwatch/handler.py
```diff
@@ -216,3 +238,61 @@ def parse_urls_txt(urls_txt):
...
+                sys.exit(1)
+
+    return jobs
+
+def create_urls_yaml(jobs, urls_yaml):
+    with os.fdopen(os.open(urls_yaml, os.O_WRONLY | os.O_CREAT, 0o0644), 'w') as f:
```
thp: Why not just `with open(urls_yaml, 'w') as f:`?

lib/urlwatch/handler.py
```diff
@@ -216,3 +238,61 @@ def parse_urls_txt(urls_txt):
...
+        for job in jobs:
+            if isinstance(job, ShellJob):
+                f.write("---\n")
```
thp: Also please don't manually generate YAML. Generate a Python structure and then use the YAML module (that you already use for parsing) for generating the YAML. This takes care of proper escaping, etc...

urlwatch
```diff
@@ -53,9 +53,11 @@ line_length = 75
 # File and folder paths
 import sys
 import os.path
+from os import rename
```
thp: As above - use `import os` here and `os.rename` below.

urlwatch
```diff
@@ -243,19 +252,30 @@ if __name__ == '__main__':
...
-        if os.path.exists(urls_txt_example) and not os.path.exists(urls_txt_fn):
-            shutil.copy(urls_txt_example, urls_txt_fn)
+        if os.path.exists(urls_yaml_example) and not os.path.exists(urls_yaml_fn):
+            shutil.copy(urls_yaml_example, urls_yaml_fn)
         if not options.hooks and os.path.exists(hooks_py_example) and not os.path.exists(hooks_py_fn):
             shutil.copy(hooks_py_example, hooks_py_fn)
         sys.exit(1)
 
+    if os.path.isfile(urls_txt):
+        log.warning('Old urls.txt found, but ignored.')
```
thp: Maybe log as:

    log.warning('Ignoring old urls.txt file at %s', urls_txt)

This way, in case the user is unsure where the program found the file, they have an easier time removing it.

thp:
Added some more comments. Needs a bit more work, but we're getting there :)

lib/urlwatch/handler.py
egroeper: Because then the file could be group writable (depending on systems umask settings) and urlwatch would stop with a security warning, if the file contains ShellJobs.
I don't want to create a file automatically, that makes urlwatch stop working (because of permission problems).

lib/urlwatch/handler.py
egroeper: I understand your intention. Using the library definitely results in a valid file. But yaml has many correct format alternatives. It's hard to configure the style the way I want (and think is readable for the users). I think I now found a way, that's good enough.

I would prefer, if the string values would be quoted (especially for shell commands). But when changing "default_style" the attribute names get quoted, too. That doesn't match my taste...

egroeper:
Did another round.

lib/urlwatch/handler.py
```diff
@@ -216,3 +238,59 @@ def parse_urls_txt(urls_txt):
...
+                    print >>sys.stderr, '\n  Please remove shell jobs or fix these problems.\n'
+                    sys.exit(1)
+            else:
+                log.warning("Entry has neither url nor command! Ignoring: %r", job)
+
+    return jobs
+
+def create_urls_yaml(jobs, urls_yaml):
+    yaml_jobs = []
+    with os.fdopen(os.open(urls_yaml, os.O_WRONLY | os.O_CREAT, 0o0644), 'w') as f:
```
thp: `with open(urls_yaml, 'w') as fp:`

egroeper: No problem. It's your coude.
But this happens on my Ubuntu with default umask settings:
```
enno@dharma ~/git/urlwatch[extendable_urls_file*]$ cat ~/.urlwatch/urls.txt

# This is an example urls.txt file for urlwatch
# Empty lines and lines starting with "#" are ignored

http://www.dubclub-vienna.com/
http://www.openpandora.org/developers.php
#http://www.statistik.tuwien.ac.at/lv-guide/u107.369/info.html
#http://www.statistik.tuwien.ac.at/lv-guide/u107.369/blatter.html
#http://www.dbai.tuwien.ac.at/education/dbs/current/index.html
#http://www.dbai.tuwien.ac.at/education/dbs/current/uebung.html
http://ti.tuwien.ac.at/rts/teaching/courses/systems_programming
http://ti.tuwien.ac.at/rts/teaching/courses/systems_programming/labor
http://ti.tuwien.ac.at/rts/teaching/courses/betriebssysteme
#http://www.complang.tuwien.ac.at/anton/lvas/effiziente-programme.html
#http://www.complang.tuwien.ac.at/anton/lvas/effizienz-aufgabe08/
http://www.kukuk.at/ical/events
http://guckes.net/cal/

# You can use the pipe character to "watch" the output of shell commands
|ls -al ~

# If you want to use spaces in URLs, you have to URL-encode them (e.g. %20)
http://example.org/With%20Spaces/

# You can do POST requests by writing the POST data behind the URL,
# separated by a single space character. POST data is URL-encoded.
http://example.com/search.cgi button=Search&q=something&category=4

enno@dharma ~/git/urlwatch[extendable_urls_file*]$ ll ~/.urlwatch/
insgesamt 24
drwxrwxr-x 2 enno enno 4096 Dez 12 01:16 cache
drwxrwxr-x 2 enno enno 4096 Okt  4 13:16 lib
-rw-r--r-- 1 enno enno 1218 Dez 12 01:04 urls.txt
-rw-rw-r-- 1 enno enno 1218 Okt  4 13:16 urls.txt.example
-rw-rw-r-- 1 enno enno 1514 Okt  4 13:22 urls.yaml.example
enno@dharma ~/git/urlwatch[extendable_urls_file*]$ gvim lib/urlwatch/handler.py
enno@dharma ~/git/urlwatch[extendable_urls_file*]$ ./urlwatch 

Found old urls.txt and imported it into /home/enno/.urlwatch/urls.yaml

An example for urls.yaml is available in /home/enno/.urlwatch/urls.yaml.example
Old urls.txt has been renamed to urls.txt.imported


You can also create /home/enno/.urlwatch/lib/hooks.py
An example is available in /home/enno/.urlwatch/lib/hooks.py.example

enno@dharma ~/git/urlwatch[extendable_urls_file*]$ ll ~/.urlwatch/
insgesamt 28
drwxrwxr-x 2 enno enno 4096 Dez 12 01:16 cache
drwxrwxr-x 2 enno enno 4096 Okt  4 13:16 lib
-rw-rw-r-- 1 enno enno 1218 Okt  4 13:16 urls.txt.example
-rw-r--r-- 1 enno enno 1218 Dez 12 01:04 urls.txt.imported
-rw-rw-r-- 1 enno enno  528 Jan  3 19:26 urls.yaml
-rw-rw-r-- 1 enno enno 1514 Okt  4 13:22 urls.yaml.example
enno@dharma ~/git/urlwatch[extendable_urls_file*]$ ./urlwatch 

  SECURITY WARNING - Cannot run shell jobs:

     /home/enno/.urlwatch/urls.yaml is group/world-writable

  Please remove shell jobs or fix these problems.

enno@dharma ~/git/urlwatch[extendable_urls_file*]$
```

Because of this I added this more complex code, that doesn't have that issue.
But of course in most cases this only happens once and the code would stay longer. It's a decision you as the maintainer have to make. I only want you to know all the facts.
I will apply your patch. I could revert it, if you change your mind.

egroeper:
Applied your patch (see my line note) and fixed my warning outputs.

I'm quite unsure if this is a good way to emit the warnings.
They are now only shown in verbose mode and I needed the additional log parameter in my function calls.
But simply writing to stderr by hand doesn't sound like a good idea, either.

thp:
I have created #32 on top of your changes and cleaned up the code a bit more, so it's easier to maintain. I have taken your commits and squashed it to a single commit there, but kept your authorship information, then modified it so it's easier for me to maintain it in the future. As a bonus, command line arguments for managing jobs have been added (--list, --add, --delete and --edit).

Please have a look what you think and if it's fine with you, we can close this one and merge the other one.

thp:
Merged as #32 now.

Issue #21, opened 2015-07-20
Filter diff results


csill1634:
It is possible to define custom hooks to "prepare" the content of the website for parsing. However, this happens before calculating the diff. I'm wondering if there is an easy way to define hooks that are executed after calculating the diff? For example, sometimes I only want to be informed if content is added to an website and do not care about the content that has been removed since the last visit. The simplest way (as I'm doing now with a quickly hacked grep parser) would be to remove any lines starting with "-" from the resulting diff. Is there any way to achieve this cleanly with urlwatch?

thp:
Hm, in general you couldn't distinguish between stuff being added and stuff being modified (without a more advanced diff). However, I'd be happy to accept a patch that optionally allows for "diff" hooks that take both inputs and return a diff (overriding the default diff algorithm). This way, modifications could be detected as well as new additions.

thp:
See also #25

thp:
In urlwatch 2, diffing is part of a "Reporter" class, and you can write your own (the output of that can be something that's not even a diff, and also it can either go to stdout or sent via e-mail, etc...

PR #25, opened 2015-11-14
Support for using custom external diff command

wrobelda:
This allows for using more robust,  content-aware comparison tools, e.g. for XML, JSON, etc.

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,10 +63,10 @@ Q: What about badly-formed HTML (long lines, etc..)?
 A: Use python-utidylib. See the example hooks.py file.
 
 Q: Is there a way to compare the data using a different diff tool (eg. syntax-aware XML diff)?
-A: Use "--diff-tool-command 'command'", where command is a combination of full filesystem path
-   to the binary tool and its arguments. Substitute compared files' names with '{OLD_DATA}' and
-   '{NEW_DATA}' tokens. This doesn't support pipelining multiple commands - put these in a sh
-   script and use this instead.
+A: Use "--diff-command=COMMAND", where COMMAND is a full filesystem path to the
+   tool followed by its arguments. Substitute compared files' names with '{OLD_DATA}'
+   and '{NEW_DATA}' tokens. This doesn't support pipelining of multiple
+   commands - put these in a sh script and use this instead.
 
 Q: Is there a way to make the output more human-readable?
 Q: Is there a way to turn it into a diff of parsed HTML perhaps?
--- a/README.md
+++ b/README.md
@@ -62,6 +62,12 @@ A: Indeed there is. See the example hooks.py file.
 Q: What about badly-formed HTML (long lines, etc..)?
 A: Use python-utidylib. See the example hooks.py file.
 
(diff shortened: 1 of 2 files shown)


urlwatch
```diff
@@ -370,6 +394,19 @@ if __name__ == '__main__':
             if display_errors:
                 details += foutput('error', job, (repr(error) +
                         '\n' + str(error)).strip(), summary)
+        except OSError, error:
+            log.error('got OSError while trying to execute diff command: %s' % error)
```
thp: The `log.error()` function does `%`-formatting by itself, so `log.error('got ... %s', error)` would be better here (and below).

urlwatch
```diff
@@ -370,6 +394,19 @@ if __name__ == '__main__':
...
+            if display_errors:
+                details += foutput('error', job, error, summary)
+        except KeyError, error:
+            log.error('got error while attempting to replace diff command\'s placeholder(s): %s' % error)
+            if display_errors:
+                details += foutput('error', job, error, summary)
+        except subprocess.CalledProcessError, error:
+            log.error('got error while executing diff command %s: %s' % (error.cmd, error.output))
```
thp: Here it would be `log.error('got error while executing diff command %s: %s', error.cmd, error.output)`

README.md
```diff
@@ -62,6 +62,12 @@ A: Indeed there is. See the example hooks.py file.
 Q: What about badly-formed HTML (long lines, etc..)?
 A: Use python-utidylib. See the example hooks.py file.
 
+Q: Is there a way to compare the data using a different diff tool (eg. syntax-aware XML diff)?
+A: Use "--diff-command=COMMAND", where COMMAND is a full filesystem path to the
+   tool followed by its arguments. Substitute compared files' names with '{OLD_DATA}'
+   and '{NEW_DATA}' tokens. This doesn't support pipelining of multiple
+   commands - put these in a sh script and use this instead.
```
thp: Given that we don't support pipelining (should we? subprocess has `shell=True` for that), maybe we shouldn't request `{OLD_DATA}` and `{NEW_DATA}` but just pass in the first and second argument? (and if the diff utility needs different parameter ordering and/or additional commands, just let a shell script deal with those, and call the shell script with the two positional arguments)

thp:
Added some comments. Looks like a really useful feature, wonder if overriding the diff utility should be done per-URL (e.g. I would want to use one diff utility for a XML document, and a different one for something else, e.g. word diff).

thp:
@wrobelda: Any chance to update this PR with the comments from above?

wrobelda:
Sure I will, was too busy last weeks. Will do this weekend.

thp:
Please update this so it fits the new YAML style model (maybe have a per-URL configuration option that can specify the external diff command)?

thp:
As mentioned in #21, maybe this can be implemented as new Reporter subclass for urlwatch 2?

wrobelda:
Sorry for late reply - I have been out for holidays starting mid-December. I will look into this over this weekend or following week.

thp:
Closing this for now, if you have time to update this and rebase against the current version, feel free to reopen this one or open a new pull request.

PR #30, opened 2016-01-01
added new hooks

volkan:


thp:
Please do not check in the *.pyc bytecode files.

thp:
Also, please provide a description what your change does and what the goal of the change is - i.e. what use cases there are for using those features that you add.

volkan:
Sorry for pyc files. I fixed it.

1.diff_after_hook which is used when the difference appears. For cache invalidation like varnish or other tools.
2.request_hook is used to customize main request or to add the process of a new request. Like login process.

thp:
For the "PURGE" (request hook) requests, maybe that could be a new job type after #32 has been merged? Also, on top of #32 we could probably make the hooks a bit nicer in general (or also make a login job type that then calls some hooks).

volkan:
It sounds good.

thp:
These changes have now been merged. For a simple change in HTTP method, you can now just specify `method: PURGE` in the YAML file. For logins, you can define a new job type that handles the login. I have added corresponding examples to `urls.yaml.example` and `hooks.py.example`.

Issue #15, opened 2015-04-25
Support for cookie auth


ealprr:
Some webpage need session cookie and currently, urlwatch does not seems to handle those case.
`urllib2` need `cookielib` to handle cookies but `requests` manage this automatically.

sbraz:
Hi, I would love to see this as well.

thp:
You could prepare a patch that replaces urllib usage with requests?

sbraz:
@thp I'll look into it.

thp:
Now that requests is in, this might be easier to implement. @ealprr can you explain for which pages this happens and what it should do exactly?

sbraz:
@thp It's probably trivial to add a cookies dict to the YAML. Do you want me to make a PR for this?

thp:
Yes, if you're interested in working on this, I'd be all for it :)

PR #35, opened 2016-01-22
Replace urllib usage with requests

sbraz:
Hi @thp, let me know what you think of this as a patch for https://github.com/thp/urlwatch/issues/15
This is just a quick rewrite, I might have missed some corner cases. I decided to re-use `HTTPError` for 304 responses, since requests will not raise an exception for these.

### DIFF
--- a/lib/urlwatch/jobs.py
+++ b/lib/urlwatch/jobs.py
@@ -31,9 +31,7 @@
 import subprocess
 import re
 
-import urllib.request
-import urllib.error
-import urllib.parse
+import requests
 
 import email.utils
 import zlib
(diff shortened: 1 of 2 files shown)


thp:
Wow, nice change - this simplifies the code quite a bit :)

Just create a `NotModified` subclass of `Exception` (like `ShellError` but without any special handling) and raise that (remove the urllib usage and `raise UnmodifiedError()`), will clean up the code a bit more:

```
class NotModified(Exception):
    ...
```

And then in the retrieval function:

```
    if response.status_code == 304:
        raise NotModified()
```

Note that the body of the class can indeed be `...` (like that) in Python 3, which IMHO is a bit nicer than `pass`.

Also, add "(Fixes #15)" to the commit message, so it is automatically closed when merged.

sbraz:
OK, I made a custom exception class. It does not fix #15 entirely since that issue was about cookies.

sbraz:
Hey @thp, is this good to merge now? Have you had the chance to test it yet?

thp:
Looks good, thanks for working on this!

PR #36, opened 2016-01-27
Add cookies support

sbraz:
Fixes https://github.com/thp/urlwatch/issues/15
I'll let you document it yourself if that's alright :)


thp:
Yup!

Issue #38, opened 2016-01-28
Usee Markdown for README


sbraz:
I think the README would benefit from Markdown, here is a simple conversion from yours (I might have missed stuff).
https://gist.github.com/sbraz/4df782cb3225500c63d0

thp:
Sounds good; can you create a pull request for that?

PR #39, opened 2016-01-28
Use Markdown for README

sbraz:
https://github.com/thp/urlwatch/issues/38


sbraz:
The ASCII art doesn't look too sexy, maybe it should be dropped? I'll let you decide :)

thp:
Nah, it's cool :)

PR #41, opened 2016-01-30
Add a new auto-applying filter that uses regexes, fixes #37

### LINKED ISSUE #37: Regex support for AutoMatchFilter
This would be nice and it's something I'll miss if I stop using the legacy filters. Example use case: tracking packages from websites that always share the same base URL.

sbraz:
@thp as discussed, here's the new regex-based filter.
Can you make a new release with all the recent additions? That'd be great :)
Fixes https://github.com/thp/urlwatch/issues/37


thp:
Thanks. Sure I can do a new release; if I don't have one done by next weekend, ping me and I'll get right to it.

PR #54, opened 2016-02-22
Added support for pushover

atiro:
Another attempt, with much reduced code using the new system. Only sending it as text for the moment, as the HTML changed message doesn't seem to display well via pushover.


lib/urlwatch/reporters.py
```diff
@@ -338,3 +343,39 @@ def submit(self):
...
+    """Send summary via Pushover"""
+
+    __kind__ = 'pushover'
+
+    def submit(self):
+        cfg = self.report.config['report']['pushover']
+        filtered_job_states = list(self.report.get_filtered_job_states(self.job_states))
+
+        body_text = '\n'.join(super().submit())
+#        body_html = '\n'.join(super().submit())
```
thp: Please remove this commented-out line

lib/urlwatch/storage.py
```diff
@@ -77,6 +77,11 @@
                 'keyring': True,
             },
         },
+        'pushover': {
+            'enabled': False,
+            'app': '',
+            'user': ''
+            }
```
thp: Please fix the indentation of the closing `}` and add a `,` after it so that adding new sections makes the diff nicer.

thp:
Added some style comments. Also, please add a short section to the `README` file how to get the values for "app" and "user" in the configuration file (like I think you had previously, where to obtain those keys and what to put in, etc..

atiro:
Tidied up & added brief section to README

lib/urlwatch/reporters.py
```diff
@@ -338,3 +343,38 @@ def submit(self):
...
         mailer.send(msg)
+
+
+class PushoverReport(TextReporter):
+    """Send summary via Pushover"""
+
+    __kind__ = 'pushover'
+
+    def submit(self):
+        cfg = self.report.config['report']['pushover']
```
thp: This line can be removed.

lib/urlwatch/reporters.py
```diff
@@ -338,3 +343,38 @@ def submit(self):
...
+        if len(body_text) > 1024:
+            body_text = body_text[0:1023]
+
+        try:
+            app = chump.Application(cfg['app'])
+        except:
+            logger.error("Failed to load chump - is it installed ('pip install chump')")
+            return
+
+        user = app.get_user(cfg['user'])
```
thp: Use `self.config['user']` here

lib/urlwatch/reporters.py
```diff
@@ -338,3 +343,38 @@ def submit(self):
...
+
+        if not body_text:
+            logger.debug('Not sending pushover (no changes)')
+            return
+
+        if len(body_text) > 1024:
+            body_text = body_text[0:1023]
+
+        try:
+            app = chump.Application(cfg['app'])
```
thp: Use `self.config['app']` here

thp:
Have one more comment that I just saw now. Instead of looking up the configuration section from the global config, just use `self.config`, which assumes less about the config structure.

After that it's ready for merge.

atiro:
Updated config calls

PR #60, opened 2016-03-12
added option to use sendmail instead of SMTP

e-dschungel:


thp:
Basically good, but:

* Put the sendmail-calling code (`import subprocess` and all) into `mailer.py`, as a subclass of `Mailer` with an overridden `send()` method (if you want to make it very clean, make `Mailer` an abstract class by raising `NotImplementedError` in `send()` and have a `SMTPMailer` and `SendmailMailer` subclass that both provide only the `send()` method)
* I'd make `-t -oi` the default arguments and only make the sendmail binary configurable, with the default value being `sendmail` (i.e. look for sendmail in `$PATH`)
* Instead of checking if `stderr` contained anything, check the exit code of the binary (e.g. with `.wait()`)
* Strings are not concatenated with `.` in Python - this isn't PHP

e-dschungel:
Should be working and contains all your suggestions

thp:
Great, thanks :)

PR #67, opened 2016-04-01
Add link to watched location in email report

layus:
This is not acceptable as-is, as some lacations may not be urls (like scripts invocation...) but it gives the idea.


thp:
Do you think you could rework that (with some simple regex) to work for both URLs and non-URLs?

layus:
It would be easier to get the "type" of the watched object. If it is an url, make it a link.

thp:
Yeah, however you want to solve this so that linking only happens for URLs. Can't merge it in as-is at the moment.

layus:
Updated!

thp:
Looks good. Please remove the two semicolons at the end of the lines, they are not needed in Python and are not Python style. After that we can merge this.

layus:
Oh, oh, too much c in the meantime :-). Done!

thp:
Thanks a lot, merged :)

Issue #52, opened 2016-02-17
pip uninstall urlwatch deletes /usr/share/


Eriner:
This may have been a fluke, but in removing urlwatch with pip (to use Arch's python-urlwatch instead), pip deleted my whole `/usr/share/`. I was able to recover, but I don't want that to happen to anyone else. I don't have time to dig into exactly why it happened, but based on a conversation in another thread, there seems to be some issue with where/how urlwatch installs (and uninstalls).

brbsix:
Wow that's pretty crazy. Is there any reason in particular that you were running pip with sudo/root privileges? Assuming you aren't working in a virtualenv or something, you'd generally you'd only want to install with the `--user` flag which doesn't require elevated privileges.

Eriner:
It was in a virtual machine that I was playing with some stuff in; everything was running as root. The deletion of `/usr/share/` wasn't a big deal, I didn't even have to restore from a snapshot; a few lines in bash and I was good-to-go again. Just wanted to make the issue in case anyone else ran into this (or if this is a bug with urlwatch that can be avoided).

thp:
Pushed a change that might improve the situation. Previously, the `setup.py` script had an invisible assumption coded in that you run the setup.py script from the source directory. Now, it should work correctly even when running from somewhere else. Still unsure why it happened, but it could be that if you run `setup.py` with the current working directory being `/usr` (because it scans `share` in the current directory to see which files are to be installed, which works great when running from the source tree).

chid:
Could you please push v2.2 to pypi? This bug still occurs on that version.

thp:
Done.

PR #73, opened 2016-06-16
Fix setup.py install

sbraz:
Hi, I'm not sure exactly how this is supposed to work but the current setup.py prevents the package to install for me. `m['data_files']` looks like this:
```
[('hare/man/man1', ['hare/man/man1/urlwatch.1']), ('hare/urlwatch/examples', ['hare/urlwatch/examples/hooks.py.example', 'hare/urlwatch/examples/urls.yaml.example']), ('hare/urlwatch/examples/__pycache__', ['hare/urlwatch/examples/__pycache__/hooks.py.cpython-34.pyc', 'hare/urlwatch/examples/__pycache__/hooks.py.cpython-35.pyc'])]
```
And `setup.py install` fails with
```
error: can't copy 'hare/man/man1/urlwatch.1': doesn't exist or not a regular file
```

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -31,7 +31,7 @@ m['scripts'] = [os.path.join(HERE, PACKAGE_NAME)]
 m['package_dir'] = {'': os.path.join(HERE, 'lib')}
 m['packages'] = ['.'.join(dirname[len(HERE)+1:].split(os.sep)[1:])
                  for dirname, _, files in os.walk(os.path.join(HERE, 'lib')) if '__init__.py' in files]
-m['data_files'] = [(dirname[len(HERE)+1:], [os.path.join(dirname[len(HERE)+1:], fn) for fn in files])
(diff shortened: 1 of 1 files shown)


thp:
On which OS/Python version are you experiencing this issue?

sbraz:
Python 3.4  or 3.5 on Linux. After a bit of digging, I realised that the problem only seems to occur if I use `--root` with `setup.py install`. Even if my PR fixes the install process, I think there are prettier ways to do what you're trying to achieve (I assume it is just stripping the root from a path?).

thp:
The commit linked above should fix this. Was a bug introduced by https://github.com/thp/urlwatch/commit/50356a779ff40cd914408e7e9a34c54b08ef4c26 which in turn is a fix for #52. The initial fix was bad, we should just not allow running setup.py in any other folder than the source tree. If this breaks some packaging/build scripts, then those should be fixed.

sbraz:
Thanks! Can you release a new version with the fix?

thp:
Doing a release now, thanks!

PR #85, opened 2016-07-30
Ignore InsecureRequestWarning

sbraz:
If `ssl_no_verify` is enabled for one URL, urllib3 will issue warnings:
```
/usr/lib64/python3.5/site-packages/urllib3/connectionpool.py:790: InsecureRequestWarning: Unverified HTTPS request is being made. Adding certificate verification is strongly advised. See: https://urllib3.readthedocs.org/en/latest/security.html
  InsecureRequestWarning)
```
Since we control all the requests we make, I think we can globally silence this warning.

### DIFF
--- a/lib/urlwatch/jobs.py
+++ b/lib/urlwatch/jobs.py
@@ -43,7 +43,9 @@ import ssl
 import urlwatch
 
 from .util import TrackSubClasses
+from requests.packages.urllib3.exceptions import InsecureRequestWarning
 
+requests.packages.urllib3.disable_warnings(InsecureRequestWarning)
 
 logger = logging.getLogger(__name__)
 


thp:
The documentation at https://urllib3.readthedocs.io/en/latest/advanced-usage.html#ssl-warnings (linked from the warning message) suggests using one of the following instead:

```
>>> import urllib3
>>> urllib3.disable_warnings()
```

or

```
>>> logging.captureWarnings(True)
```

Also, I wonder if it would be better to also enable certificate verification and only disable certificate verification on a per-URL basis (via the `urls.yaml` file)?

sbraz:
> Also, I wonder if it would be better to also enable certificate verification and only disable certificate verification on a per-URL basis (via the urls.yaml file)?

That is already the case, that's why we don't need to issue warnings :) → https://github.com/thp/urlwatch/blob/d62ae5a97b85601812fee34c30de2bb471a5ee29/lib/urlwatch/jobs.py#L206

sbraz:
But yeah, logging warnings could be nice too.

thp:
Ok, makes sense. Feel free to open a new PR if you want to add logging the warnings.

sbraz:
Well on the other hand, maybe it's better to show the warnings to the
user...
Can you make a new release with the current bug fix?

PR #90, opened 2016-08-28
File job

lechuckcaptain:
Added FileJob class in order to process differences on local files (could be also useful for easier testing)

### DIFF
--- a/lib/urlwatch/jobs.py
+++ b/lib/urlwatch/jobs.py
@@ -39,6 +39,7 @@ import hashlib
 import base64
 import logging
 import ssl
+import os.path
 
 import urlwatch
 
@@ -230,3 +231,29 @@ class UrlJob(Job):
                 return response.content.decode('ascii', 'ignore')
 
         return response.text
+
+class FileJob(Job):
+    """Read the content of a local file"""
+
+    __kind__ = 'file'
+    __required__ = ('path',)
+
+    def get_location(self):
+        return str(self.path)
+
+    def retrieve(self, job_state):
+
+        path_string = str(self.path)
+
+        if not path_string:
+            raise ValueError('Path cannot be null or empty')
+
+        if path_string.startswith('file://'):
+            path_string = path_string.replace('file://', '')
+
+        if not os.path.exists(path_string):
+            raise ValueError("Cannot find file: '{0}'".format(path_string))
+
+        response = open(path_string, 'r').read()
+        return response
+
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,2 @@
 __pycache__
+.idea


lib/urlwatch/jobs.py
```diff
@@ -230,3 +231,29 @@ def retrieve(self, job_state):
...
+        return str(self.path)
+
+    def retrieve(self, job_state):
+
+        path_string = str(self.path)
+
+        if not path_string:
+            raise ValueError('Path cannot be null or empty')
+
+        if path_string.startswith('file://'):
```
thp: I wouldn't handle this here. Come to think of it, are local files not already supported by using a `file://` URL? If not, maybe they should be? This way, we don't need a new kind of job, but can just use a URL job with `file://`, something that might be more straightforward?

lechuckcaptain:
Unfortunately it seems that requests library doesn't handle automatically `file://` or `ftp://` URLs. 

The only available solution is to define a custom adapter to handle different protocols.

More information here: 

http://docs.python-requests.org/en/master/user/advanced/#transport-adapters
http://stackoverflow.com/questions/10123929/python-requests-fetch-a-file-from-a-local-url

What do you think would be the best solution?

lib/urlwatch/jobs.py
lechuckcaptain: Unfortunately it seems that requests library doesn't automatically support `file://` or `ftp://` protocols. The only available option would be to define a custom adapter to support each different protocol. More information here:

- http://docs.python-requests.org/en/master/user/advanced/#transport-adapters
- http://stackoverflow.com/questions/10123929/python-requests-fetch-a-file-from-a-local-url

What do you think would be the best solution?

thp: We don't need to "send in" the data into requests. In the line before `requests.request()` is called (currently here: https://github.com/thp/urlwatch/blob/master/lib/urlwatch/jobs.py#L206), one can do the `if url.startswith('file://'):` check there and return early. That should be simple enough to not have to deal with extending/modifying requests, while still allowing normal file:// URLs transparently for local files.

lechuckcaptain: Maybe I don't fully get your proposal, since to me It seems a kludge that won't be helpful if some user decide to use alternative protocols such as `ftp://`. Am I right? Can you try to clarify?

thp: This pull request is only for local files, adding special-casing for `file://` URIs should be enough. No need to deal with alternative protocols such as `ftp://`. URLs starting with `file://` get special treatment (we /do/ know that there's no point in giving `file://` to `requests.request()`), and everything else will be passed to requests, whether it works or not.

lechuckcaptain: OK I got it now. But in my opinion it would be a cleaner solution implementing a transport adapter, so in the future we can further extend the supported protocols. I can prepare a proposal if you want.

thp:
Simpler alternative in #111, please have a look. If you're okay with that, i'd close this one here.

PR #91, opened 2016-08-29
UrlJob proxy support

lechuckcaptain:
Added proxy support to UrlJob

### DIFF
--- a/lib/urlwatch/jobs.py
+++ b/lib/urlwatch/jobs.py
@@ -28,6 +28,7 @@
 # THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 
+import os
 import subprocess
 import re
 
@@ -182,7 +183,7 @@ class UrlJob(Job):
     __kind__ = 'url'
 
     __required__ = ('url',)
-    __optional__ = ('cookies', 'data', 'method', 'ssl_no_verify')
+    __optional__ = ('cookies', 'data', 'method', 'ssl_no_verify', 'http_proxy', 'https_proxy')
 
     CHARSET_RE = re.compile('text/(html|plain); charset=([^;]*)')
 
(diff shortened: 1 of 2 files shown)


thp:
Does this break using environment variables for proxy support? Other than that, looks good.

lechuckcaptain:
You're right! I added the read of proxy settings from os environment. The system settings can be overriden (or directly defined when is not possible to set environment variables) by the configuration added into urls.yaml file.

lib/urlwatch/jobs.py
```diff
@@ -26,8 +26,7 @@
 # THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
 # (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF
 # THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
-
```
thp: Please leave in the newlines here, looks much better.

lib/urlwatch/jobs.py
```diff
@@ -182,18 +181,24 @@ class UrlJob(Job):
...
-    __optional__ = ('cookies', 'data', 'method', 'ssl_no_verify')
+    __optional__ = ('cookies', 'data', 'method', 'ssl_no_verify', 'http_proxy', 'https_proxy')
 
     CHARSET_RE = re.compile('text/(html|plain); charset=([^;]*)')
 
     def get_location(self):
         return self.url
 
     def retrieve(self, job_state):
+
```
thp: Please no empty newlines at the start of function body.

lib/urlwatch/jobs.py
```diff
@@ -182,18 +181,24 @@ class UrlJob(Job):
...
         headers = {
             'User-agent': urlwatch.__user_agent__,
         }
 
+        proxies = {
+            'http': os.getenv('HTTP_PROXY'),
+            'https': os.getenv('HTTPS_PROXY')
```
thp: Add trailing `,` at the end of this line, it's easier to add new items to the dict this way in the future if needed.

lechuckcaptain:
Applied feedback from @thp. Should be fine now.

thp:
All good now, thanks :)

PR #103, opened 2016-10-13
Filter for HTML "class" element

strauhmanis:
Useful when HTML "id" element is not enough.

### DIFF
--- a/lib/urlwatch/filters.py
+++ b/lib/urlwatch/filters.py
@@ -287,7 +287,7 @@ class ElementByclass(html.parser.HTMLParser):
             self._result.append(data)
 
 
-class GetElementByclass(FilterBase):
+class GetElementByClass(FilterBase):
     """Get a HTML element by its class"""
 
     __kind__ = 'element-by-class'
--- a/lib/urlwatch/filters.py
+++ b/lib/urlwatch/filters.py
@@ -252,6 +252,53 @@ class GetElementById(FilterBase):
         element_by_id.feed(data)
         return element_by_id.get_html()
 
+class ElementByclass(html.parser.HTMLParser):
+    def __init__(self, element_class):
+        super().__init__()
+
+        self._element_class = element_class
+        self._result = []
+        self._inside = False
+        self._depth = 0
+
+    def get_html(self):
+        return ''.join(self._result)
+
+    def handle_starttag(self, tag, attrs):
+        ad = dict(attrs)
+
+        if ad.get('class', None) == self._element_class:
+            self._inside = True
+
+        if self._inside:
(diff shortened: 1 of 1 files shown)


lib/urlwatch/filters.py
```diff
@@ -252,6 +252,53 @@ def filter(self, data, subfilter=None):
         element_by_id.feed(data)
         return element_by_id.get_html()
 
+class ElementByclass(html.parser.HTMLParser):
```
thp: This is mostly a copy of `ElementByID` -- please refactor it so that the class is called e.g. `ElementByAttribute` and pass in `"id"` or `"class"` as constructor parameter in the `GetElementById` and `GetElementByClass` filters.

lib/urlwatch/filters.py
```diff
@@ -252,6 +252,53 @@ def filter(self, data, subfilter=None):
...
+            self._depth -= 1
+            if self._depth == 0:
+                self._inside = False
+
+    def handle_data(self, data):
+        if self._inside:
+            self._result.append(data)
+
+
+class GetElementByclass(FilterBase):
```
thp: Please rename to `GetElementByClass` (note the uppercase "C")

ad-m:
I think CSS selectors (```https://www.crummy.com/software/BeautifulSoup/bs4/doc/#css-selectors```) and ```XPath``` are most general way to parse HTML/XML. CSS selectors mostly are conversable to XPath (https://pythonhosted.org/cssselect/). It supports both select by id and select by class and a lot more different use case. Maybe it is better to use them, @strauhmanis?

strauhmanis:
Looking at the examples on the site, this seems to be a very elegant and powerful way to target a needed part of a web page. Though, besides very minor code contributions, like what I’ve requested to pull, my Python skill would not allow me to build on this idea, I’m afraid.

thp:
Both options (`ElementByAttribute` with two specializations for class and id, as well as making it more generic with CSS selectors) work for me. Waiting for @strauhmanis to update the patch to use the common superclass. If you don't have time to work on that, maybe I'll find some time to look into it at some point, let's keep this PR open for now.

strauhmanis:
Hi Thomas @thp ! The PR I put forth, is as good it will get, I'm afraid. Writing a general-purpose programming language code isn't an area of my expertise—not by a long shot. As such, you should close this PR, I suppose.

strauhmanis:
Hi Thomas! The PR I put forth, is as good it will get, I'm afraid. Writing code in a general-purpose programming language isn't an area of my expertise—not by a long shot. As such, you should close this PR, I suppose.

thp:
Superseded by #110.

Issue #113, opened 2016-11-01
A traceback


strauhmanis:
Hi Thomas! Thank you for the recent #110 (HTML class filter), first of all 👍 

When running the current Git master, there's a following output on my system (macOS 10.12.1):

urlwatch 2.5, Copyright 2008-2016 Thomas Perl
Website: http://thp.io/2008/urlwatch/
watched 30 URLs in 2 seconds
Traceback (most recent call last):
  File "/Users/ervsts/Git/urlwatch/urlwatch", line 101, in <module>
    urlwatch_command.run()
  File "/Users/ervsts/Git/urlwatch/lib/urlwatch/command.py", line 167, in run
    self.urlwatcher.close()
  File "/Users/ervsts/Git/urlwatch/lib/urlwatch/main.py", line 96, in close
    self.report.finish()
  File "/Users/ervsts/Git/urlwatch/lib/urlwatch/handler.py", line 125, in finish
    ReporterBase.submit_all(self, self.job_states, duration)
  File "/Users/ervsts/Git/urlwatch/lib/urlwatch/reporters.py", line 90, in submit_all
    subclass(report, cfg, job_states, duration).submit()
  File "/Users/ervsts/Git/urlwatch/lib/urlwatch/reporters.py", line 445, in submit
    result = requests.post(
NameError: name 'requests' is not defined

I don't quite know what's wrong here.

Thank you!

ad-m:
@strauhmanis , do you use mailgun reporter? Do you can provide your config file?

ad-m:
@lechuckcaptain , did you miss ```import requests```?

strauhmanis:
Hi Adam @ad-m ! I'm not sure what you mean about missing import requests, sorry about me not being very technical minded, but I do send email notifications about changes via mailgun service.

Below are contents of my urlwatch.yaml:

display:
  error: true
  new: true
  unchanged: false
report:
  email:
    enabled: false
    from: 'myemail@gmaill.com'
    html: false
    method: smtp
    sendmail:
      path: sendmail
    smtp:
      host: localhost
      keyring: true
      port: 25
      starttls: true
    subject: '{count} changes: {jobs}'
    to: 'myemail@gmaill.com'
  html:
    diff: unified
  mailgun:
    api_key: 'key-xxxx'
    domain: 'sandboxbxxxx.mailgun.org'
    enabled: true
    from_mail: 'myemail@gmaill.com'
    from_name: 'urlwatch'
[...]

PR #114, opened 2016-11-01
Add missing requests import for mailgun

ad-m:
See #113

### DIFF
--- a/lib/urlwatch/reporters.py
+++ b/lib/urlwatch/reporters.py
@@ -50,6 +50,11 @@ try:
 except ImportError:
     Pushbullet = None
 
+try:
+    import requests
+except ImportError:
+    requests = None
+
 logger = logging.getLogger(__name__)
 
 


sbraz:
@ad-m Why would you not display an error if requests is not found? urlwatch requires it anyway.

ad-m:
@sbraz , I proportioned to the style of old code (see https://github.com/ad-m/urlwatch/blob/dee9868055a17cf4c1ca104d30fce2ebd89d7ffa/lib/urlwatch/reporters.py#L49 ).

sbraz:
Well pushbullet is optional, requests isn't. Anyway, no big deal :)

ad-m:
requests is required only by ``MailGunReporter``. pushbullet is required only by ``PushbulletReport``. I don't see any differences. Using both of reporters are not required.

sbraz:
Requests is _not_ optional. urlwatch depends on it: https://github.com/thp/urlwatch/blob/master/setup.py#L16

ad-m:
@sbraz , you are right.

PR #128, opened 2017-02-03
Fix email encodings

raifthenerd:
Some mail clients (e.g. Microsoft Mail) cannot display UTF-8 properly due to the wrong encoding labels, so i fixed.
(ref: https://encoding.spec.whatwg.org/#names-and-labels)


thp:
Thanks :)

PR #130, opened 2017-02-04
Update handler.py

yantar92:
Allow filters to have more than one subfilter like filter:param1:param2


thp:
Why do you use `re.split()`? And is there any handler that would make use of multiple subfilters? Also the commit message is useless, of course you have updated handler.py -- please add a proper commit message why these changes are done, and also update the pull request title (same goes for the other pull requests).

thp:
Closing this, as there has been no feedback and it's not clear to me what is to be achieved by this pull request. Feel free to reopen with a better description.

PR #131, opened 2017-02-04
Update handler.py

yantar92:
Ignore escaped ',' and ':' in filter


thp:
Please merge this into #130.

PR #133, opened 2017-02-05
Add XPath expression support

ptitsy:
Added a filter that returns elements matching an XPath expression.

Note that by switching to Parsel (https://github.com/scrapy/parsel), it is also entirely possible to replace GetElementById and GetElementsByClass with a universal CSS selector.


thp:
Good change, I'm happy to merge that. Just a question, there's XPath support in the standard library [xml.etree.ElementTree](https://docs.python.org/3.6/library/xml.etree.elementtree.html) module, do you think that could be used for XPath resolving and the additional dependency avoided?

thp:
@ptitsy Did you have time to look at my previous comment? Can the `parsel` dependency be replaced with `xml.etree.ElementTree`'s XPath support? Do we have a test case / example to verify XPath functionality?

thp:
@ptitsy ping ^ see my last comment

thp:
@ptitsy ^ ping again

thp:
I'd personally prefer CSS selectors over XPath.

thp:
Maybe something like this: https://cssselect.readthedocs.io/en/latest/

PR #135, opened 2017-02-24
Telegram reporter

gvdenbro:
adding documentation for the telegram configuration and added chunking of messages bigger than 4096 characters (message limit of telegram).


thp:
Very good, thanks!

PR #136, opened 2017-03-02
Adding support of url that contains comma.

afaucon:
The command:
`urlwatch --add url=http://www.mysite.com,name=MySite`
does not work if mysite is:
[http://www.foo.com/page?&zips=1400,1410&orderby=date](http://www.foo.com/page?&zips=1400,1410&orderby=date)
because in the implementation you have supposed that there is no comma in the url.
The issue is raised by this line in command.py:
`d = {k: v for k, v in (item.split('=', 1) for item in self.urlwatch_config.add.split(','))}`
Supposing there is a higher probability to have a comma in the url than in the name, we could use instead:
`d = {k: v for k, v in (item.split('=', 1) for item in self.urlwatch_config.add.rsplit(',', 1))}`

### DIFF
--- a/lib/urlwatch/command.py
+++ b/lib/urlwatch/command.py
@@ -130,7 +130,7 @@ class UrlwatchCommand:
                     save = False
 
         if self.urlwatch_config.add is not None:
-            d = {k: v for k, v in (item.split('=', 1) for item in self.urlwatch_config.add.split(','))}
+            d = {k: v for k, v in (item.split('=', 1) for item in self.urlwatch_config.add.rsplit(',', 1))}
             job = JobBase.unserialize(d)
             print('Adding %r' % (job,))
             self.urlwatcher.jobs.append(job)


thp:
Did you try putting `%2C` into the URL (which is `,` URL-encoded)? See also: http://stackoverflow.com/a/8828831/1047040

Does this work for you?

    urlwatch --add url=http://www.foo.com/page?&zips=1400%2C1410&orderby=date,name=MySite

We can't just do a maxsplit, since the `--add` command allows 1..n key/value pairs, it's not restricted to exactly two.

Note that another workaround would be to add a dummy URL first, and then use the `--edit` command to open the URL list YAML in a text editor (where you can add the comma as-is to the URL).

afaucon:
Oh yes, it works using %2C.

And I understand that --add is not restricted to exactly two kay/value pairs. 
Hence my pull request can be closed without accepting my modifications.

PR #137, opened 2017-03-08
Change HTTP header Content-Type for POST requests

vmassuchetto:
This is a solution for the general case when using the `data` option, as some sites will only allow POST requests to form results with this specification.

However, another approach would be creating a new `headers` option to allow other `Content-type` values (`application/json` for APIs, for example), but I don't know if that's the case.


thp:
That's probably okay. If we have breakage due to this, we can always make the content-type configurable as you say.

PR #139, opened 2017-03-15
Option to avoid 304 responses

vmassuchetto:
Some websites will return cached responses even when they get changed. This implements the `ignore_cached` option so the GET header is adjusted to avoid 304 response and the body text is always fetched.


thp:
Merged, thanks.

PR #141, opened 2017-03-18
ElementsByAttribute: look for matching tag in handle_endtag

Cryptosaurus:
The ElementsByAttribute filter just counts opening and close tags to detect the end of the matched element.  Unfortunately, this sometimes fails, even with valid HTML, because some end tags are optional; for instance, it fails with `<ul><li>Foo<li>Bar</ul>`.

This patch maintains a list of nested elements, and handle_starttag pops elements untill a matching one is found. It seems to work, but I haven't tested exhaustively...

### DIFF
--- a/lib/urlwatch/filters.py
+++ b/lib/urlwatch/filters.py
@@ -216,7 +216,7 @@ class ElementsByAttribute(html.parser.HTMLParser):
         self._attributes = {name: value}
         self._result = []
         self._inside = False
-        self._depth = 0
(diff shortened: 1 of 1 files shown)


thp:
Looks good, thanks. Let's see if there are any corner cases that are not detected and add fixes then.

akurtz:
Excellent and crucial fix for this filter -- thanks, @Cryptosaurus.

Any chance this can get into a point release?

PR #142, opened 2017-03-22
Option to determine a minimum time interval for jobs

vmassuchetto:
This implements the `interval` option for jobs. It gets a human-readable string and will skip the job for this time since it was last processed.

    url: http://some-url.com
    interval: 3 hours

So it's possible to schedule urlwatcher for, let's say, every 5 minutes in the crontab, and it will fetch each URL at the configured frequency.


vmassuchetto:
There's a problem on variable checks. Will review it and submit later.

snowman:
reminder from me, please consider review it and submit, it's 2020 now :)

PR #145, opened 2017-04-06
html2text method to prevent links from breaking

vmassuchetto:
Adds a `pyhtml2text_nolinebreak` method that will prevent links from getting line breaks.

### DIFF
--- a/lib/urlwatch/html2txt.py
+++ b/lib/urlwatch/html2txt.py
@@ -60,6 +60,13 @@ def html2text(data, method='lynx'):
         d = pyhtml2text.handle(data)
         return d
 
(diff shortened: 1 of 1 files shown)


vmassuchetto:
Please see #146

lib/urlwatch/html2txt.py
```diff
@@ -60,6 +60,13 @@ def html2text(data, method='lynx'):
         d = pyhtml2text.handle(data)
         return d
 
+    if method == 'pyhtml2text_nolinebreak':
```
thp: Please add this also to the docstring above, otherwise this feature is not discoverable.

vmassuchetto:
#146 covers this.

Issue #146, opened 2017-04-07
Make subfilter parameters configurable


vmassuchetto:
For any given set of different method parameters, we would have to create a new function to support it. As it's the case with https://github.com/thp/urlwatch/pull/145 for just one simple extra configuration.

Maybe it would be better if we make it configurable, for example:

    filter: html2text:pyhtml2text:body-width=0:protect-links=1:images-with-size=1
    filter: html2text:lynx:list_inline=1:unique_urls=1

Is that a good idea? I would be happy to implement it.

thp:
Yes, although in this case instead of "inventing" a syntax for filter configuration, maybe the `filter` element in the YAML can be either a string (to have the legacy behavior) or a more structured element, that can then have some nicer syntax (i.e. key = filter name, value = a dict of key, value pairs for configuration):

```
filter:
- html2text:
    body-width: 0
    method: pyhtml2text
    protect-links: true
```

So: The filter key has either a string value (then handle as with legacy code), or it has a list value, in which case each list element has as key the filter name and as value the filter configuration (dict). List value so that one can chain different filters (like `,` in the legacy filter string) one after the other.

PR #150, opened 2017-05-29
Filters in YAML and configurable html2text

vmassuchetto:
This fixes thp/urlwatch#146, implements YAML lists for filters and
reads configurable parameters for the html2text filters

### DIFF
--- a/lib/urlwatch/filters.py
+++ b/lib/urlwatch/filters.py
@@ -78,6 +78,7 @@ class FilterBase(object, metaclass=TrackSubClasses):
 
     @classmethod
     def process(cls, filter_kind, subfilter, state, data):
+        logger.info('Applying filter %r, subfilter %r to %s', filter_kind, subfilter, state.job.get_location())
         filtercls = cls.__subclasses__.get(filter_kind, None)
         if filtercls is None:
             raise ValueError('Unknown filter kind: %s:%s' % (filter_kind, subfilter))
@@ -155,11 +156,18 @@ class Html2TextFilter(FilterBase):
     __kind__ = 'html2text'
 
     def filter(self, data, subfilter=None):
-        if subfilter is None:
-            subfilter = 're'
 
+        if subfilter is None:
+            method = 're'
+            options = {}
+        elif isinstance(subfilter, dict):
+            method = subfilter.pop('method')
+            options = subfilter
+        elif isinstance(subfilter, str):
+            method = subfilter
+            options = {}
         from .html2txt import html2text
-        return html2text(data, method=subfilter)
+        return html2text(data, method=method, options=options)
 
 
 class Ical2TextFilter(FilterBase):
--- a/lib/urlwatch/html2txt.py
+++ b/lib/urlwatch/html2txt.py
@@ -36,7 +36,7 @@ import logging
 logger = logging.getLogger(__name__)
 
 
-def html2text(data, method='lynx'):
+def html2text(data, method='lynx', options=None):
 
     """
     Convert a string consisting of HTML to plain text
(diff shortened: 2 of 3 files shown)


thp:
Please fix the style errors that are reported by Travis CI:

```
/home/travis/build/thp/urlwatch/lib/urlwatch/filters.py:82:17: E128 continuation line under-indented for visual indent
/home/travis/build/thp/urlwatch/lib/urlwatch/html2txt.py:89:37: E231 missing whitespace after ','
```

lib/urlwatch/filters.py
```diff
@@ -155,11 +157,18 @@ class Html2TextFilter(FilterBase):
     __kind__ = 'html2text'
 
     def filter(self, data, subfilter=None):
-        if subfilter is None:
-            subfilter = 're'
 
+        if subfilter is None:
+            method = 're'
+            options = dict()
```
thp: Use `{}` to create a dictionary.

lib/urlwatch/filters.py
```diff
@@ -155,11 +157,18 @@ class Html2TextFilter(FilterBase):
...
+        elif isinstance(subfilter, dict):
+            method = subfilter.pop('method')
+            options = subfilter
+        elif isinstance(subfilter, str):
+            method = subfilter
+            options = dict()
```
thp: Use `{}` to create a dictionary.

lib/urlwatch/html2txt.py
```diff
@@ -36,18 +36,23 @@
 logger = logging.getLogger(__name__)
 
 
-def html2text(data, method='lynx'):
+def html2text(data, method='lynx', options=dict()):
```
thp: This doesn't work as expected, you must use `options=None` here and then:

    if options is None:
        options = {}

See also: https://stackoverflow.com/questions/26320899/why-is-the-empty-dictionary-a-dangerous-default-value-in-python

lib/urlwatch/html2txt.py
```diff
@@ -56,25 +61,33 @@ def html2text(data, method='lynx'):
...
+        cmd = ['lynx', '-nonumbers', '-dump', '-stdin', '-assume_charset UTF-8', '-display_charset UTF-8']
     elif method == 'html2text':
         cmd = ['html2text', '-nobs', '-utf8']
-        stdout_encoding = 'utf-8'
     else:
         raise ValueError('Unknown html2text method: %r' % (method,))
 
+    stdout_encoding = 'utf-8'
+
+    if len(options) > 0:
```
thp: There is no point in doing this check, since iterating over an empty dict will do nothing. Also there is no point in using `iter(options)` and then using `options[o]`, instead you can do this and it should be nicer to read and better optimizable by Python:

```
for k, v in options.items():
    cmd.append(...)
```

Also if you want to check if the value is true, do just that (`== True`), not `isinstance(options[o], bool)`, as this will do something unexpected for `{'some_option': False}` (the option will be set anyway).

lib/urlwatch/html2txt.py
```diff
@@ -56,25 +61,33 @@ def html2text(data, method='lynx'):
 
     if method == 'pyhtml2text':
         import html2text
-        pyhtml2text = html2text.HTML2Text()
-        d = pyhtml2text.handle(data)
+        parser = html2text.HTML2Text()
+        if len(options) > 0:
+            for o in iter(options):
```
thp: Same here, the `if` and `for` can be combined into:

```
for k, v in options.items():
    setattr(parser, k.lower(), v)
```

vmassuchetto:
Thanks for the corrections. Applied in in 34127cce84410897ce980d1de49399071d47b5d4.

The main idea in `iter(options)` was to modify its contents while iterating, but this ended up being unnecessary in the logic.

thp:
Thanks, this is much nicer now.

PR #157, opened 2017-06-26
Allow 'user' parameter for SMTP sending

jsit:


thp:
Clean & easy, thank you very much :)

Issue #42, opened 2016-02-03
Don't create .urlwatch if not necessary


kyrias:
I move the config, url list, hooks file, and cache to other directories because I don't like having my home directory cluttered with dotdirectories unnecessarily, (consider implementing the XDG basedir spec instead..,) but urlwatch insists on creating the `.urlwatch` directory even if it isn't actually used.

thp:
Yes, that's a good idea. It should use XDG on freedesktop.org-like systems, ~/Library/Application Support/ on OS X and %APPDATA% on Windows, but for compatibility with existing installations use ~/.urlwatch/ if it exists and the new location does not exist (if both exist, it can use the new location). New installations will also use the new location.

jwilk:
You might want to use [appdirs](https://pypi.python.org/pypi/appdirs) to implement this.

kyrias:
Normally I'd say to just implement it yourself, since it's a tiny amount of code, but I guess if you want to support Windows it gets a lot more annoying to have to deal with.

PR #166, opened 2017-09-05
Add XDG_CONFIG_DIR support

jelly:
Add support for XDG_CONFIG_DIR by adding a dependency to appdir while
keeping compatibility with the existing configuration location.

Closes: #42


lib/urlwatch/filters.py
```diff
@@ -130,6 +132,9 @@ def __init__(self, job, state):
         super().__init__(job, state)
 
         self.hooks = None
+        if not os.path.exists(self.FILENAME):
+            self.FILENAME = AppDirs('urlwatch').user_config_dir + '/lib/hooks.py'
```
thp: Use `os.path.join()` instead of hardcoding `/` as path separator.

share/man/man1/urlwatch.1
```diff
@@ -71,13 +71,13 @@ list supported jobs/filters/reporters
 remove old cache entries
 .SH "FILES"
 .TP
-.B ~/.urlwatch/urls.yaml
+.B ~/.config/urlwatch/urls.yaml
```
thp: Well, this should mention `$XDG_CONFIG_DIR`.

lib/urlwatch/filters.py
jelly: Fixed

share/man/man1/urlwatch.1
jelly: Okido replaced it with XDG_CONFIG_DIR :)

lib/urlwatch/filters.py
```diff
@@ -130,6 +132,9 @@ def __init__(self, job, state):
         super().__init__(job, state)
 
         self.hooks = None
+        if not os.path.exists(self.FILENAME):
+            self.FILENAME = os.path.join(AppDirs('urlwatch').user_config_dir, 'lib/hooks.py')
```
thp: I meant:

    self.FILENAME = os.path.join(AppDirs('urlwatch').user_config_dir, 'lib', 'hooks.py')

urlwatch
```diff
@@ -36,9 +36,14 @@ import signal
 import socket
 import sys
 
+from appdirs import AppDirs
+
 pkgname = 'urlwatch'
 urlwatch_dir = os.path.expanduser(os.path.join('~', '.' + pkgname))
 
+if not os.path.exists(urlwatch_dir):
+    urlwatch_dir = AppDirs('urlwatch').user_config_dir
```
thp: Use `pkgname` instead of `'urlwatch'`

lib/urlwatch/filters.py
thp: Or actually, wait -- legacy hooks would only be in the older folder, so there's no need to change anything here.

thp: Meaning if it doesn't exist, don't try to find it in the XDG-based directory.

urlwatch
jelly: Ah of course, fixed that.

lib/urlwatch/filters.py
jelly: Reverted it all to the original.

Issue #155, opened 2017-06-14
check body tag only?


kunwon1:
Any way to check only the body tag? I can't find a filter to handle this

ad-m:
There is no build-in filters for that. It will be good to use xpath for that. Here is PR to add XPath  #133. Unmerged yet, unfortunately. Do you can fix that PR to add support to XPath and check body tag only?

PR #169, opened 2017-09-13
Add possibility to filter for tag

cmichi:
I had the use case of having to restrict urlwatch to the body and this is the way I realized it.

This PR would also solve #155.

### DIFF
--- a/lib/urlwatch/filters.py
+++ b/lib/urlwatch/filters.py
@@ -36,6 +36,8 @@ import imp
 import html.parser
 import hashlib
 
+from enum import Enum
+
 from .util import TrackSubClasses
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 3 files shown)


cmichi:
Good point. I have amended the PR. Can you please take a look again?

lib/urlwatch/filters.py
```diff
@@ -217,11 +219,21 @@ def filter(self, data, subfilter=None):
         return data.strip()
 
 
-class ElementsByAttribute(html.parser.HTMLParser):
-    def __init__(self, name, value):
+class FilterBy(Enum):
+    attribute = 1
```
thp: Make those named constants uppercase (`ATTRIBUTE = 1` and `TAG = 2`)

lib/urlwatch/filters.py
```diff
@@ -217,11 +219,21 @@ def filter(self, data, subfilter=None):
...
+    tag = 2
+
+
+class ElementsBy(html.parser.HTMLParser):
+    def __init__(self, filterBy, name, value=None):
```
thp: PEP-8: `filterBy` -> `filter_by`

cmichi:
Alright, I have updated the PR with your comments.

cmichi:
Thank you as well! You could close #155 now too.

PR #180, opened 2017-10-05
Retry on connection error

### LINKED ISSUE #143: Ignore transient network errors.
I run urlwatch on my kimsuffi server (OVH). About twice in a week, I get notified of network errors.

```
HTTPSConnectionPool(host='api.github.com', port=443): Max retries exceeded with url: /repos/xxx/yyy/commits/master (Caused by NewConnectionError('<requests.packages.urllib3.connection.VerifiedHTTPSConnection object at 0x7f6ed424d710>: Failed to establish a new connection: [Errno -2] Name or service not known',))
```

These errors are transient, and occur for all my urls at the same time (not just github as above).

Could we add a delay before reporting such issues ? I would like to get notifications only if connections errors occur three times in a row.

cmichi:
This PR fixes #143, which I also experience occasionally. It implements a retry mechanism which retries on network errors such as `ConnectionResetError` or `NewConnectionError`.

### DIFF
--- a/lib/urlwatch/jobs.py
+++ b/lib/urlwatch/jobs.py
@@ -146,7 +146,7 @@ class JobBase(object, metaclass=TrackSubClasses):
 
 class Job(JobBase):
     __required__ = ()
-    __optional__ = ('name', 'filter')
+    __optional__ = ('name', 'filter', 'max_tries')
 
     def pretty_name(self):
         return self.name if self.name else self.get_location()
--- a/test/data/invalid-url.yaml
+++ b/test/data/invalid-url.yaml
@@ -1,0 +1,4 @@
+name: "invalid url"
+url: "https://invalid"
+max_tries: 2
+---
--- a/test/test_handler.py
+++ b/test/test_handler.py
@@ -139,3 +139,76 @@ def test_unserialize_with_unknown_key():
         'unknown_key': 123,
         'name': 'hoho',
     })
+
+
+def prepare_retry_test():
+    urls = os.path.join(os.path.dirname(os.path.dirname(__file__)), 'test', 'data', 'invalid-url.yaml')
+    config = os.path.join(os.path.dirname(__file__), 'data', 'urlwatch.yaml')
+    cache = os.path.join(os.path.dirname(__file__), 'data', 'cache.db')
+    hooks = ''
+
+    config_storage = YamlConfigStorage(config)
+    cache_storage = CacheMiniDBStorage(cache)
+    urls_storage = UrlsYaml(urls)
+
+    urlwatch_config = TestConfig(config, urls, cache, hooks, True)
+    urlwatcher = Urlwatch(urlwatch_config, config_storage, cache_storage, urls_storage)
+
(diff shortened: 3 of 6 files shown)


lib/urlwatch/jobs.py
```diff
@@ -187,6 +188,25 @@ def get_location(self):
         return self.url
 
     def retrieve(self, job_state):
+        if hasattr(self, 'tries') is False:
```
thp: Use `not hasattr()` instead of `hasattr() is False`

thp: Or even nicer using a local variable:

    tries = getattr(self, 'tries', 3)

lib/urlwatch/jobs.py
```diff
@@ -187,6 +188,25 @@ def get_location(self):
...
+            self.tries = 3
+        if hasattr(self, 'retry_pause') is False:
+            self.retry_pause = 20
```
thp: Not sure if we should by default have a 20 second (* 3 retries) pause. This could really prolong the runtime of the script. Can you make it opt-in instead (only 1 retry by default, or having a config-global default value and being able to override per job)? Also, usually back-off is implemented using exponential timeouts (e.g. 2, 4, 8, 16 seconds) -- see https://en.wikipedia.org/wiki/Exponential_backoff

test/test_handler.py
```diff
@@ -139,3 +143,24 @@ def test_unserialize_with_unknown_key():
         'unknown_key': 123,
         'name': 'hoho',
     })
+
+
+@raises(ConnectionError)
+@with_setup(teardown=teardown_func)
+def test_connection_exception_is_thrown_after_retries_failed():
+    tries = 3
```
thp: Come to think of it, `retries` would be a better name than `tries`.

maximbaz:
Not all transient errors are network errors, for example I'm watching a few Github links and occasionally Github responds with `(500 or 502 or 503) Server Error`.

I'd like to propose a different implementation: instead of retrying in place (and thus delaying script execution), do one attempt (as today), but in case of any error alert user only if a count of consequent errors exceeds some threshold. 

We can store this error count per URL somewhere in the database, increment on error and reset on a successful request. Then alert the user only when the error count exceeds a configured threshold. 

By default you can have the threshold equal to `1`, so that the default behavior stays exactly as today - one error happened and user is notified. But then some of us can customize the threshold to say `3`, and receive an email only if the request failed 3 times consequently, i.e. `urlwatch` was executed 3 times and every time a request to this URL returned an error, any error.

@thp, @cmichi what do you think about it?

maximbaz:
This is awesome @cmichi, thank you very much! Looking forward to see this merged 😛

test/data/invalid-url.yaml
```diff
@@ -0,0 +1,4 @@
+name: "invalid url"
+url: "https://invalid"
```
thp: OK as per RFC6761.

thp:
Merged, thanks!

PR #181, opened 2017-10-11
utf8 encoding

przemas75:
Without this, it does not work with some non ascii chars.


thp:
You just have to set your system locale to UTF-8, then it will work without this patch.

PR #188, opened 2017-12-02
avoid exception if there is no ['smtp']['user'] field in config

e-dschungel:
avoid exception
```
if self.config['smtp']['user']:
KeyError: 'user'
```
if there is no ['smtp']['user'] field in config, like it is the case for older existing configs.

### DIFF
--- a/lib/urlwatch/reporters.py
+++ b/lib/urlwatch/reporters.py
@@ -343,11 +343,8 @@ class EMailReporter(TextReporter):
             logger.debug('Not sending e-mail (no changes)')
             return
         if self.config['method'] == "smtp":
-            if 'user' in self.config['smtp'] and self.config['smtp']['user']:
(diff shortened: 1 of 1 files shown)


lib/urlwatch/reporters.py
```diff
@@ -343,7 +343,7 @@ def submit(self):
             logger.debug('Not sending e-mail (no changes)')
             return
         if self.config['method'] == "smtp":
-            if self.config['smtp']['user']:
+            if 'user' in self.config['smtp'] and self.config['smtp']['user']:
```
thp: Maybe instead:

    smtp_user = self.config['smtp'].get('user', self.config['from'])

e-dschungel: Yeah, thats cleaner

PR #194, opened 2018-01-16
Correct XDG env variable for configuration

jelly:
appdirs.user_config_dir points to the location in XDG_CONFIG_HOME and
not XDG_CONFIG_DIR, incorrectly introduced with the addition of the
appdirs.user_config_dir change.


thp:
Thanks!

PR #195, opened 2018-01-16
Loading urlwatch file with a non-relative path

mwerlen:
Hi,

I'm trying to package urlwatch-2.7 as a deb package and nose tests are always in error.
It seems to come from the "lib.urlwatch" import. This modification permits pybuild (debian build utility) to run successfully nose tests.

Is this change acceptable (I'm no python expert) ?

Maxime

### DIFF
--- a/test/test_handler.py
+++ b/test/test_handler.py
@@ -11,10 +11,10 @@ import tempfile
 import os
 import imp
 
-from lib.urlwatch import storage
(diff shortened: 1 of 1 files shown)


thp:
Yes, this should be good. In case Travis CI fails, I'll fix up the `$PYTHONPATH` there.

PR #208, opened 2018-02-24
Adds fallback on editor when $VISUAL and $EDITOR are absent

mwerlen:
urlwatch allow to edit config files with an urlwatch command. This command call an
editor for this purpose. Editor is chosen by reading $VISUAL and $EDITOR. If both are
absent then urlwatch now attemps to call editor


thp:
As far as I know, `/usr/bin/editor` is a Debian-ism, coming from the `dpkg` package (`man update-alternatives`). As such, it's probably better to keep this patch in the Debian packaging only, as it won't be that useful on other systems that don't have `/usr/bin/editor`.

PR #217, opened 2018-03-11
Moving cache file to XDG_CACHE_DIR instead or XGD_CONFIG_DIR

mwerlen:
Here is a pull request to change default cache directory from XGD_CONFIG_DIR to XDG_CACHE_DIR.

Three cases are handled:
1.) If the file exists in XDG_CONFIG_HOME, use that one
2.) If the file doesn’t exist, create a new one in XDG_CACHE_HOME
3.) If both files exist, either fail or use the XDG_CACHE_HOME file (presumably newer)

Man page has been updated


thp:
Thanks!

PR #240, opened 2018-05-17
Make the bs4 html2text subfilter actually strip HTML tags

sbraz:
Currently, bs4 is just used to prettify the HTML.


thp:
Merged, thanks!

PR #249, opened 2018-06-10
.gitignore: Ignore common files in a Python project

kbabioch:
This adds some common files in a Python project to the .gitignore file.
These were mostly taken from GitHub's own `gitignore` repository [1]
[2] and make sure that metadata created during running and testing is
not interfering with the actual content.

[1]: https://github.com/github/gitignore/blob/master/Python.gitignore
[2]: https://github.com/github/gitignore/blob/master/Global/Tags.gitignore

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,67 @@
-__pycache__
+# IDEs
 .idea
-build
+
+# Byte-compiled / optimized / DLL files
+__pycache__/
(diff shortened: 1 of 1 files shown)


thp:
Not going to take this in for now, as it's too broadly defined. Let's only add files to `.gitignore` as they are created in the normal build/development process.

PR #250, opened 2018-06-17
Json format filter

nikobockerman:
Add json format filter. The implementation is originally from @gvdenbro but that change was never pushed to the official repo. I only modified it a little bit in the second commit.

### DIFF
--- a/lib/urlwatch/filters.py
+++ b/lib/urlwatch/filters.py
@@ -35,6 +35,7 @@ import os
 import imp
 import html.parser
 import hashlib
+import json
 
 from enum import Enum
 
@@ -189,12 +190,10 @@ class JsonFormatFilter(FilterBase):
     __kind__ = 'format-json'
 
     def filter(self, data, subfilter=None):
-        from json import dumps
-        from json import loads
-        parsed_json = loads(data)
+        parsed_json = json.loads(data)
         if subfilter is not None:
             parsed_json = parsed_json[subfilter]
-        return dumps(parsed_json, sort_keys=True, indent=4)
+        return json.dumps(parsed_json, sort_keys=True, indent=4)
 
 
 class GrepFilter(FilterBase):
--- a/lib/urlwatch/filters.py
+++ b/lib/urlwatch/filters.py
@@ -186,7 +186,7 @@ class Ical2TextFilter(FilterBase):
 class JsonFormatFilter(FilterBase):
     """Convert to formatted json"""
 
-    __kind__ = 'formatted-json'
+    __kind__ = 'format-json'
 
     def filter(self, data, subfilter=None):
         from json import dumps
--- a/lib/urlwatch/filters.py
+++ b/lib/urlwatch/filters.py
@@ -35,7 +35,6 @@ import os
 import imp
 import html.parser
 import hashlib
-from textwrap import indent
 
 from enum import Enum
 
@@ -187,14 +186,14 @@ class Ical2TextFilter(FilterBase):
 class JsonFormatFilter(FilterBase):
     """Convert to formatted json"""
 
-    __kind__ = 'formatted_json'
+    __kind__ = 'formatted-json'
 
     def filter(self, data, subfilter=None):
         from json import dumps
         from json import loads
-        if subfilter is None:
-            return dumps(loads(data), sort_keys=True, indent=4)
-        parsed_json = loads(data)[subfilter]
+        parsed_json = loads(data)
+        if subfilter is not None:
+            parsed_json = parsed_json[subfilter]
         return dumps(parsed_json, sort_keys=True, indent=4)
 
 
--- a/lib/urlwatch/filters.py
+++ b/lib/urlwatch/filters.py
@@ -35,6 +35,7 @@ import os
 import imp
 import html.parser
 import hashlib
+from textwrap import indent
 
 from enum import Enum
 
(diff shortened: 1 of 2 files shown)


nikobockerman:
To explain further the need for this JsonFormatFilter: Many backend calls are nowadays implemented with XHR requests that return Json formatted data for the web page to display. And usually that data is densely formatted so that it only contains one line. With this filter that data gets pretty printed and it will be easier to see from the diff what part of the data actually changed.

lib/urlwatch/filters.py
```diff
@@ -183,6 +183,20 @@ def filter(self, data, subfilter=None):
         return ical2text(data)
 
 
+class JsonFormatFilter(FilterBase):
+    """Convert to formatted json"""
+
+    __kind__ = 'formatted-json'
```
thp: Maybe just call this `format-json` to be more in line with the "verb"-style filter names (what does it do?) vs the "noun"-style filter name (what is the output?).

test/test_filters.py
```diff
@@ -35,3 +36,22 @@ def test_get_element_by_tag_nested():
     """, 'div')
     print(result)
     eq_(result, """<div>foo</div><div>bar</div>""")
+
+
+def test_json_format_filter():
+    json_format_filter = JsonFormatFilter(None, None)
+    result = json_format_filter.filter(
+        """{"field1": {"f1.1": "value"},"field2": "value"}""")
+    print(result)
```
thp: No need for a print here.

test/test_filters.py
```diff
@@ -35,3 +36,22 @@ def test_get_element_by_tag_nested():
...
+        result,
+        """{\n    "field1": {\n        "f1.1": "value"\n    },\n    "field2": "value"\n}"""
+    )
+
+
+def test_json_format_filter_subfilter():
+    json_format_filter = JsonFormatFilter(None, None)
+    result = json_format_filter.filter(
+        """{"field1": {"f1.1": "value"},"field2": "value"}""", "field1")
+    print(result)
```
thp: No need for a print here.

test/test_filters.py
```diff
@@ -35,3 +36,22 @@ def test_get_element_by_tag_nested():
...
+    eq_(result, """{\n    "f1.1": "value"\n}""")
```
thp: You should be able to write it like this, which might be easier to read and maintain later on:
```
    eq_(result, """{
    "f1.1": "value"
}""")

File: test/test_filters.py [hunk 5]
```diff
@@ -35,3 +36,22 @@ def test_get_element_by_tag_nested():
...
     print(result)
     eq_(result, """<div>foo</div><div>bar</div>""")
+
+
+def test_json_format_filter():
+    json_format_filter = JsonFormatFilter(None, None)
+    result = json_format_filter.filter(
+        """{"field1": {"f1.1": "value"},"field2": "value"}""")
+    print(result)
+    eq_(
```
thp: Also this could be written as:
```
    eq_(result, """{
    "field1": {
        "f1.1": "value"
    },
    "field2": "value"
}""")
```

lib/urlwatch/filters.py
```diff
@@ -183,6 +183,20 @@ def filter(self, data, subfilter=None):
...
+
+    def filter(self, data, subfilter=None):
+        from json import dumps
```
thp: Since the `json` module ships with Python 3, please do the `import json` at the top of the file, and use `json.loads()` / `json.dumps()` here.

lib/urlwatch/filters.py
```diff
@@ -183,6 +183,20 @@ def filter(self, data, subfilter=None):
...
+        from json import loads
+        parsed_json = loads(data)
+        if subfilter is not None:
```
thp: That's kind of undocumented, wonder if it would make sense to configure e.g. indentation with the subfilter instead, and make the getting some keyed item from JSON a separate filter instead of tacking it onto the "formatting" filter (IMHO it might be counter intuitive that formatting also "processes" the JSON).

test/test_filters.py
nikobockerman: I put this print here because all existing filter tests already have print in the same place. When test fails there isn't any way to see what the result actually contains unless it's printed here. And when tests are run with nosetests it only outputs these prints when tests fail. So I do think it makes sense to keep this print here.

lib/urlwatch/filters.py
nikobockerman: Great idea. I didn't really like the idea of filtering one item from the json especially as it only allows filtering it from the root level. I changed it now to allow specifying the indentation level.

lib/urlwatch/filters.py
nikobockerman: Done. I didn't really understand why some filters had imports in the body and some imports were at the top of the file. Thanks for giving a plausible explanation for that practise :)

lib/urlwatch/filters.py
nikobockerman: `format-json` sounds much better and it's also shorter to write. Changed.

test/test_filters.py
nikobockerman: This is definitely easier format. I already failed with the oneline format when changing this test to have different indentation. With this multi-line format it is easier to see what I'm writing as the expected formatting.

nikobockerman:
I think I now applied all the changes you requested except removing the print statements from the filter tests for which I provided a reasoning in the review comment. What do you think now of the state of this pull request?

thp:
Merged, thanks!

Issue #248, opened 2018-06-07
Add support for HTTP ETag


kbabioch:
Currently there is an `ignored_cached` option to deal with servers that return different content unexpectedly. However there seems to be no support for the ETag. urlwatch should support it, since most modern webservers and public APIs offer this functionality.

This would be mainly used for the url watching functionality (`UrlJob`), but could also be useful for features like proposed in #246. After having played around with this for a while, I'm convinced that this can be a huge bandwith and performance saver.

I'm happy to work on this, although I need some guidance. To my understanding of the code and database structure this would require a change to the database, e.g. an etag column, which would contain the etag of the last request and could be send along with any new request.

thp:
This is now merged.

PR #255, opened 2018-07-11
Implement ETag support for UrlJob

kbabioch:
This series implements ETag support in the `UrlJob` class as discussed in #248. It adds an additional column to the caching database, saves the ETag returned from the server after an initial request, and appends this ETag to any further requests. The server will then return a 304 (`Not Modified`) status code, if nothing has changed, which the existing code already handles fine (due to the support for `If-Modified-Since`).


kbabioch:
Will re-submit, once tests are fixed.

PR #256, opened 2018-07-11
Implement ETag support

kbabioch:
This series implements ETag support for `UrlJob` as discussed in #248. It adds an additional column to the caching database, saves the ETag returned from the server after an initial request, and appends this ETag to any further requests. The server will then return a 304 (`Not Modified`) status code, if nothing has changed, which the existing code already handles fine (due to the support for `If-Modified-Since`).

### DIFF
--- a/lib/urlwatch/handler.py
+++ b/lib/urlwatch/handler.py
@@ -51,6 +51,7 @@ class JobState(object):
         self.exception = None
         self.traceback = None
         self.tries = 0
+        self.etag = None
 
     def load(self):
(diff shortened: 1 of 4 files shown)


kbabioch:
I haven't tested any migration, so I can't say anything about it. There seem to be no tests for it in the suite anyway, so am I right in assuming that this needs to be tested manually?

Regarding the test case: I will look into it, but after a quick glance, there seem to be no dedicated storage tests, so I'm not sure on how to implement a ETag specific test here.

thp:
Merged, thanks :) (eventually we should add storage tests, good catch..).

Issue #245, opened 2018-06-06
Warning when content is empty


kbabioch:
Due to a misconfiguration of my filters, the actual content for some jobs was empty. I haven't realized this for a while, so I've missed some change(s). I think it is safe to assume that there is a misconfiguration in case of empty content, so there should be a warning.

In case someone wants to actually monitor empty pages, there might be some additional attribute and/or filter to disable this warning and to let urlwatch know that you expect empty content for a particular job. It should **not** be the default, though.

thp:
There is a `--test-filter` option that can be used to test the filters.

mentalstring:
I agree with @kbabioch that it would be useful to have an option (perhaps per job?) to be warned when result of the job is empty. The `--test-filter` requires testing it manually on the command line, but doesn't help when an existing working job stops being able to capture something (eg: an `element-by-id` that no longer exists on the page).

thp:
It might be a per-job option (feel free to create a pull request). In some cases there might also be a situation where an empty filter output is expected (watching for some element to appear on a webpage, for example).

PR #257, opened 2018-07-11
Raise exception in case of empty job output

kbabioch:
This addresses #245 and makes sure an exception is raised (resulting in
an error) when the output from a job is empty, since this is a pretty
good indication for a misconfiguration. The `ignore_empty` disables this
check, making it still possible to monitor empty pages, e.g. in case you
want to watch for an empty page getting filled with content.

### DIFF
--- a/lib/urlwatch/handler.py
+++ b/lib/urlwatch/handler.py
@@ -34,7 +34,7 @@ import time
 import traceback
 
 from .filters import FilterBase
-from .jobs import NotModifiedError
+from .jobs import NotModifiedError, EmptyJobOutputError
 from .reporters import ReporterBase
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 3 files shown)


kbabioch:
I can certainly add this key to `__optional__` in `Job`, but given that childs of this class are supposed to overwrite it, I'm not sure whether this is a good idea. Wouldn't this be lost for all childs then? Or should we rewrite the `__init()__` also?

Also I'm not sure if reverting the logic is useful. Nobody is going to add an `fail_if_empty` option to each entry in his configuration, so in the end nobody will use / profit from this feature.

Personally I ran into the situation of empty results quite a couple of times and sometimes didn't even realize it. I would argue that an empty output is a **strong** indication for a misconfiguration. If you desire to monitor empty pages (which should be the exception rather than the rule), you are supposed to add an option to your configuration.

As you are the maintainer, this is your decision/call, and I won't argue too much about it. Just wanted to bring up the reasoning of why I find such a feature useful.

thp:
`TrackSubClasses` should take care of merging the optional and required fields.

thp:
@kbabioch Any updates on this? Have you checked `TrackSubClasses`?

Issue #218, opened 2018-03-13
question: how to disable notifications when network is broken


monperrus:
Sometimes, crontab starts urlwatch while network is stopped or intermittent.

I then get irrelevant notifications:

    HTTPConnectionPool(host='foo.se', port=80): Max retries exceeded with url: /ddgf/ (Caused by NewConnectionError('<urllib3.connection.HTTPConnection object at 0x7fd66c79b128>: Failed to establish a new connection: [Errno 101] Network is unreachable',))

How to disable notifications when there is a `NewConnectionError`?

kbabioch:
This is currently not possible. urlwatch works under the assumption that the network is available, otherwise it makes not much sense to monitor URLs, which are mostly online resources, without network connectivity. You can fix this in the code (e.g. look at the handler/worker logic), but no with the configuration.

monperrus:
Thanks for your answer. Would you accept a pull-request adding support for this?

kbabioch:
I'm not the maintainer (only a contributor) of this project, so it's not up to me ;). @thp is probably willing to accept a PR, if it makes sense (TM). Personally, I would make this configurable via a `ignore_network` switch. It can be implemented very similarly to a recent pull request of me (#257). 

You can add `ignore_network` as an optional argument to the `UrlJob` and check for it in the handler/worker logic. Through `job_state.exception` you can evaluate what exception was raised.

kbabioch:
I've implemented this. See commit 7c19a6acf1e410251958074c21d3ce0416d70775. Please give it a try, and let me know if this solves your issue. I will submit this as pull request then.

monperrus:
`curl https://github.com/thp/urlwatch/commit/7c19a6acf1e410251958074c21d3ce0416d70775.patch | git apply`

works for me!

Thanks a lot.

thp:
#261 is merged.

PR #261, opened 2018-07-16
Add optional option to ignore network errors in UrlJob

kbabioch:
This implements a per-entry configuration option named `ignore_network`
that can be specified if (temporary) network errors should be ignored.
When this option is specified, you won't be notified in case of a
network error. The default is still the same, i.e. if this key is not
present, you'll still get an error and will be notified. This was
discussed and tested in #218.


thp:
I'll rename it post-merge, hope that's okay with you :)

PR #270, opened 2018-08-11
Avoid a crash when a Telegram user has only one name (first_name) set.

m3phisto:
The "last_name" property of a telegram chat might not be set, if not specified by the Telegram user. This leads to a KeyError, which was not caught.


thp:
See the Travis logs for code style issues:

```
/home/travis/build/thp/urlwatch/lib/urlwatch/command.py:208:29: E226 missing whitespace around arithmetic operator
/home/travis/build/thp/urlwatch/lib/urlwatch/command.py:208:42: E226 missing whitespace around arithmetic operator
```

thp:
Nevermind, that issue is actually in master ;)

PR #291, opened 2018-10-24
reset nonzero `job_state.tries` for unchanged jobs

### LINKED ISSUE #290: `job_state.tries` not reset after a successful connection if there's no change
https://github.com/thp/urlwatch/blob/2d4e513d38d279cbbc09cf367aef78bafa39b0fc/lib/urlwatch/worker.py#L70-L72

`job_state.tries` is not reset on every successful connection. This limits the usefulness of the `max_tries` parameter for web pages with unreliable connections but rarely changing content.

cfbao:
close #290

### DIFF
--- a/lib/urlwatch/handler.py
+++ b/lib/urlwatch/handler.py
@@ -91,7 +91,6 @@ class JobState(object):
                             subfilter = None
                         data = FilterBase.process(filter_kind, subfilter, self, data)
             self.new_data = data
-            self.tries = 0
 
         except Exception as e:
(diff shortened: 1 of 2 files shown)


cfbao:
Oh, just realized this doesn't fully close #290 yet.
The problem exist for non-changing page even when 304 is not returned.
I'll add another commit

cfbao:
Just pushed another commit. Now all successful & unchanged jobs should have their `tries` count reset.

PR #296, opened 2018-10-25
use colorama to support ANSI color codes on Windows

### LINKED ISSUE #295: Stdout colors not working on Windows
Windows console does not support ANSI color codes natively.

cfbao:
close #295

### DIFF
--- a/lib/urlwatch/reporters.py
+++ b/lib/urlwatch/reporters.py
@@ -323,6 +323,10 @@ class StdoutReporter(TextReporter):
         return self._incolor(4, s)
 
     def submit(self):
+        if sys.platform == 'win32':
+            import colorama
+            colorama.init()
+
         cfg = self.report.config['report']['text']
         line_length = cfg['line_length']
 
--- a/lib/urlwatch/reporters.py
+++ b/lib/urlwatch/reporters.py
@@ -305,8 +305,15 @@ class StdoutReporter(TextReporter):
 
     __kind__ = 'stdout'
 
+    def __init__(self, report, config, job_states, duration):
+        super().__init__(report, config, job_states, duration)
+        self._has_color = sys.stdout.isatty() and self.config.get('color', False)
+        if sys.platform == 'win32' and self._has_color:
+            import colorama
+            colorama.init()
+
     def _incolor(self, color_id, s):
-        if sys.stdout.isatty() and self.config.get('color', False):
+        if self._has_color:
             return '\033[9%dm%s\033[0m' % (color_id, s)
         return s
 
@@ -323,10 +330,6 @@ class StdoutReporter(TextReporter):
         return self._incolor(4, s)
 
     def submit(self):
-        if sys.platform == 'win32':
-            import colorama
-            colorama.init()
-
         cfg = self.report.config['report']['text']
         line_length = cfg['line_length']
 
--- a/lib/urlwatch/reporters.py
+++ b/lib/urlwatch/reporters.py
@@ -305,12 +305,16 @@ class StdoutReporter(TextReporter):
 
     __kind__ = 'stdout'
 
+    # make sure colorama.init() is only called once by this class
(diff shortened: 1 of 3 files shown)


cfbao:
Thanks! Windows dependencies added in `setup.py`.

lib/urlwatch/reporters.py
```diff
@@ -305,8 +305,15 @@ class StdoutReporter(TextReporter):
 
     __kind__ = 'stdout'
 
+    def __init__(self, report, config, job_states, duration):
+        super().__init__(report, config, job_states, duration)
+        self._has_color = sys.stdout.isatty() and self.config.get('color', False)
+        if sys.platform == 'win32' and self._has_color:
+            import colorama
+            colorama.init()
```
thp: Looking at the code in https://github.com/tartley/colorama/blob/master/colorama/initialise.py#L23, it looks like the `colorama` module doesn't support calling `init()` multiple times (well, at least it will probably re-wrap the streams or something, and "uniniting" might no longer be possible).

I've filed https://github.com/tartley/colorama/issues/205 in colorama, maybe it can/should be fixed there (too?).

For now (and to support old installations of the `colorama` module), maybe add a global/static member `_colorama_inited` and only import and call `colorama.init()` once.

cfbao: I've added a static member `_colorama_inited` as requested. But I'm afraid it's a bit futile given this flaw in colorama.

I noticed that, even without this patch, it's possible to get proper colored stdout report on Windows by just having a browser job, if colorama is already installed. This is because, `BrowserJob` imports `requests_html` which imports `pyppeteer` which imports `tqdm` which imports `colorama` (if available) and calls `init()`.

I really don't know how to properly handle this situation, or if it's even meaningful to handle it, when any package in urlwatch's dependency tree can potentially call `colorama.init()`, or user may write `hooks.py` that calls it. 

Earlier related issues: tartley/colorama#145, tartley/colorama#128, tartley/colorama#182

thp: Super dirty idea: We call `colorama.init()` outside of the reporter, somewhere in urlwatch's main function (before any other code like imports or hooks gets run) and then we monkey-patch the module to have a no-op init function?

```
colorama.init()

def dummy_init(*args, **kwargs):
    ...

colorama.init = dummy_init
```

cfbao: Hmmm... dirty it is.
This works for the current situation. However, it wouldn't work if imported modules do something like `from colorama import init`.

cfbao: Oh... of course it works for `from colorama import init`. What was I thinking...

But then what do we do about `colorama.deinit`? What if external code called `init` (nothing happens) then called `deinit` (color gone), after that `StdoutReporter` starts working and we have no color and a mess of ANSI codes! (Windows only of course)
Or... what if someone uses `urlwatch` as a library, then the main function has no effect... Aggghhh!

How about we don't call `init` and don't change `sys.stdout`, and just directly use colorama.AnsiToWin32 inside of `StdoutReporter.submit` (like one example in `colorama`'s README)?
This way, other modules can do whatever they like, we just ensure that `StdoutReporter.submit` has the right color, and no side effect on `sys.stdout`.

thp:
Ready to be merged once the merge conflict (due to merging of the `enum34` change) is resolved.

thp:
Something like that (assuming thread safety is not an issue, otherwise protect it with ` threading.Lock`):
```
# The original init function
def real_init():
    ...
# The original deinit function
def real_deinit():
    ...

_init_counter = 0
def init():
    global _init_counter
    if _init_counter == 0:
        real_init()
    _init_counter += 1

def deinit():
    global _init_counter
    _init_counter -= 1
    assert _init_counter >= 0, 'deinit() called without calling init()'
    if _init_counter == 0:
        real_deinit()
```

PR #298, opened 2018-10-26
render browser jobs' web pages

### LINKED ISSUE #293: 'browser' job does not render with javascript
https://github.com/thp/urlwatch/blob/2d4e513d38d279cbbc09cf367aef78bafa39b0fc/lib/urlwatch/jobs.py#L288-L292
`response.html.render()` is needed before returning `response.html.html`

cfbao:
fix #293

### DIFF
--- a/lib/urlwatch/jobs.py
+++ b/lib/urlwatch/jobs.py
@@ -282,11 +282,15 @@ class BrowserJob(Job):
 
     __required__ = ('navigate',)
 
+    _session = None
+
     def get_location(self):
         return self.navigate
 
     def retrieve(self, job_state):
-        from requests_html import HTMLSession
-        session = HTMLSession()
-        response = session.get(self.navigate)
+        # all instances of BrowserJob share the same session
+        if not BrowserJob._session:
+            from requests_html import HTMLSession
+            BrowserJob._session = HTMLSession()
+        response = BrowserJob._session.get(self.navigate)
         return response.html.html
--- a/lib/urlwatch/jobs.py
+++ b/lib/urlwatch/jobs.py
@@ -282,15 +282,21 @@ class BrowserJob(Job):
 
     __required__ = ('navigate',)
 
+    main_thread = True
     _session = None
 
     def get_location(self):
         return self.navigate
 
     def retrieve(self, job_state):
+        """
+        Retrieve web pages rendered in Chromium with JavaScript.
+        Must be called sequentially from the main thread
+        """
         # all instances of BrowserJob share the same session
         if not BrowserJob._session:
             from requests_html import HTMLSession
             BrowserJob._session = HTMLSession()
         response = BrowserJob._session.get(self.navigate)
+        response.html.render()
         return response.html.html
--- a/lib/urlwatch/jobs.py
+++ b/lib/urlwatch/jobs.py
@@ -148,6 +148,9 @@ class Job(JobBase):
     __required__ = ()
     __optional__ = ('name', 'filter', 'max_tries', 'diff_tool')
 
+    # job will run sequentially in the main thread if set to true
+    MAIN_THREAD = False
+
     def pretty_name(self):
         return self.name if self.name else self.get_location()
 
@@ -282,7 +285,7 @@ class BrowserJob(Job):
 
     __required__ = ('navigate',)
 
-    main_thread = True
+    MAIN_THREAD = True
     _session = None
 
     def get_location(self):
--- a/lib/urlwatch/worker.py
+++ b/lib/urlwatch/worker.py
@@ -41,64 +41,75 @@ logger = logging.getLogger(__name__)
 MAX_WORKERS = 10
 
 
-def run_parallel(func, items):
-    executor = concurrent.futures.ThreadPoolExecutor(max_workers=MAX_WORKERS)
-    for future in concurrent.futures.as_completed(executor.submit(func, item) for item in items):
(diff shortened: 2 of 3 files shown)


cfbao:
I added a class attribute `main_thread` to `BrowserJob`, which is used by `worker.run_jobs` to pick them out.

The idea is that users can also set this attribute in their custom classes if they want their jobs to run in the main thread sequentially.

cfbao:
Just a note that browser jobs requires Python 3.6+, as this is what requests-html requires. Earlier Python versions are unsupported by requests-html and causes errors.

cfbao:
Conflicts resolved. However, new minor conflict will arise once #303 is merged.

I suggest using "hide whitespace changes" in diff settings for a clearer view of changes.

lib/urlwatch/jobs.py
```diff
@@ -148,6 +148,9 @@ class Job(JobBase):
     __required__ = ()
     __optional__ = ('name', 'filter', 'max_tries', 'diff_tool')
 
+    # job will run sequentially in the main thread if set to true
+    MAIN_THREAD = False
```
thp: `MAIN_THREAD` -> `RUN_ON_MAIN_THREAD`

lib/urlwatch/jobs.py
```diff
@@ -282,11 +285,21 @@ class BrowserJob(Job):
...
+        """
+        Retrieve web pages rendered in Chromium with JavaScript.
+        Must be called sequentially from the main thread
+        """
+        # all instances of BrowserJob share the same session
+        if not BrowserJob._session:
+            from requests_html import HTMLSession
+            BrowserJob._session = HTMLSession()
+        response = BrowserJob._session.get(self.navigate)
+        response.html.render()
```
thp: Is is actually important that the requests are run on the main thread, or is it just important that there are no requests happening at the same time? If it's the latter, maybe a simple [`threading.Lock`](https://docs.python.org/3/library/threading.html#lock-objects) would do the trick as well without needing to change `worker.py`?

lib/urlwatch/worker.py
```diff
@@ -41,64 +41,75 @@
...
 
 
-def run_parallel(func, items):
+def run_jobs(urlwatcher):
+    cache_storage = urlwatcher.cache_storage
+    jobs = urlwatcher.jobs
+    report = urlwatcher.report
+
+    logger.debug('Processing %d jobs', len(jobs))
+    main_thread_jobs = [job for job in jobs if job.MAIN_THREAD]
```
thp: These two lines might be pre-destined for https://stackoverflow.com/q/12720151/1047040 (any of the solutions, not necessarily mine ;), but maybe it's okay as-is, your choice :)

cfbao:
### Is running on the main thread is necessary?

Because only the main thread handles signals, Chromium browser can only be launched (by [pyppeteer.launch()](https://miyakogi.github.io/pyppeteer/reference.html#pyppeteer.launcher.launch)) on the main thread. This is done on the first call to either [session.browser](https://github.com/kennethreitz/requests-html/blob/47b2ad86cab49167fb15cce787efca74c4de9fdd/requests_html.py#L725) or `session.html.render()`. After the browser is launched, it's only important that no `session.html.render()` are called at the same time (because all `render` calls of all sessions share the same event loop), but not necessarily on the main thread.

One way to not change `worker.py`, is to call `session.browser` in `BrowserJob.__init__()` to launch the browser, then use `threading.Lock` to space out all the `render()` calls. This solution relies on the fact that currently all jobs are instantiated on the main thread. My concern, however, is that this code organization hides the fact some code must be run on the main thread, which may come up as a surprise in future maintenance.

### The problem with cookies

In this patch, all `BrowserJob` instances share the same session and the same browser instance. This is good for performance, about 25% faster than launching different browser instances. However, it has the problem that different jobs' cookies can affect one another, which probably isn't what users want. To solve this problem, we can either sacrifice some performance, or...

### Is requests-html needed at all?

`requests_html` offers very little control over the browser instance. It also has a whole layer of HTTP requests and HTML handling that isn't needed or meaningful for `BrowserJob` and urlwatch in general.

For example, `requests_html` needs to first make a HTTP request itself, before calling render, which _reloads_ the web page again. This is not efficient at all. Any parameters that may be passed to `requests` would be entirely lost when reloading in browser. Even the url (`navigate`) may change after `requests` has done its job. It also doesn't support async API yet, and is relatively slow in development.

I think we'd be better off directly using [pyppeteer](https://github.com/miyakogi/pyppeteer). Its interface is entirely async (better performance). We can use [incognito browser context](https://miyakogi.github.io/pyppeteer/reference.html#pyppeteer.browser.Browser.createIncognitoBrowserContext) to solve the cookies problem mentioned earlier. We can also offer more options to users, such as which Chromium/Chrome installation to use, where to put the user data folder, etc.

### Final remarks/questions

* Should I just rewrite `BrowserJob` completely directly using `pyppeteer`?

* In any case, _some_ code must be run on the main thread. Is running them in `BrowserJob.__init__()` acceptable (so we don't change `worker.py`), or should we make this fact more explicit to the rest of the code (so `worker.py` probably needs some change).

thp:
Since `requests-html` also requires Python 3.6, I guess using `pyppeteer` directly is fine, so feel free to change that. If possible, just keep the `worker.py` changes to a minimum.

thp:
Let's go with #310 instead.

PR #301, opened 2018-10-28
allow `--features` and `--edit-*` to run without urls file

cfbao:


cfbao:
a side note/question:
Should `--gc-cache` also be allowed to run without urls file? Not sure about that.

thp:
> Should `--gc-cache` also be allowed to run without urls file? Not sure about that.

I think we should allow it, imagining this situation:

1. Somebody has URLs and has run it
2. They want to start over, so they clean the URLs list
3. They then want to run the GC cache function to start with a clean state

In this case, running gc-cache without a URLs file should be allowed. Of course, it might also be that they accidentally moved the urls file out of place and then accidentally run gc-cache, and the entries are removed, but this is a case for user's backups :)

cfbao:
Done!

PR #303, opened 2018-10-29
Use hyperlink in HtmlReporter for browser jobs

cfbao:
Add `location_is_url` to `BrowserJob` and `UrlJob`.
Use this attribute to determine whether to use hyperlink 'a' tag in HtmlReporter.
Custom job classes can also utilize this attribute.


cfbao:
Requested changes made.

I still prefer adding a static variable rather than choosing directly on `__kind__`, as the former behavior is inheritable and configurable by subclasses. This is especially relevant for subclasses in `hooks.py`, when users can't/don't want to study and modify code in the main package.

PR #305, opened 2018-11-01
Enhance Pushover support

nscaife:
These changes add the ability to individually push reports to Pushover.  I regularly go over the message size limit when I have multiple sites that update in the same batch. Additionally, Pushover inconsistently converts the URL in the message title to a clickable link, so it now populates the Pushover URL field.

I replaced the Pushover library, because chump makes excessive and unnecessary API calls during object instantiation (karanlyons/chump#12).  This issue causes a failure when sending multiple reports, and could cause the user or IP to be banned.


karanlyons:
Hey, upstream dev of that Pushover wrapper here. Just wanted to pop in and give my apologies for the issues you ran into. The library would validate user tokens on User object instantiation (which also was the only way to get their available devices from the API at the time I wrote it) with the expectation that this was both a normal usage of the API and additionally that the User objects would likely be cached in some manner.

I’ve been made aware of the API changes regarding rate limiting this endpoint (amongst many other changes) and have a fix staged pending some more information from Pushover.

You’re under no obligation to continue using chump over alternatives (and indeed I’m no longer using Pushover and haven’t received any feedback on that repo in years, so the alternatives may have more eyes on them), but regardless I did want to let you know from one dev to another that I’m sorry for trouble it gave you.

karanlyons:
Updates to chump that should avoid the issues your having are at https://github.com/karanlyons/chump/pull/13 if you’d like to test them. I just need to make some other changes to the docs, etc. before I can bundle it into a release.

I don’t know much about this project, but I have a note regarding chump’s integration that may or may not be valid: `web_service_get()` implies that the pushover application and user are getting instantiated on every call, which I think would be unnecessary inside of a long living process unless `self.config[‘app']` or `self.config['user']` changes. It’d be better practice to cache the object if so.

Unfortunately even with the changes in the above PR you’ll still run into issues with Pushover’s API if you don’t cache the `Application` & `User` object, in part due to my unwillingness to let you send a bad request down the wire. The issue specifically is that Pushover does not return any error when attempting to send a message to a device that does not exist, but instead will just silently send a message to all the user’s devices (this is, I believe, the only parameter in their API that silently coerces like this). There is no way to get a list of user’s devices except for calling /1/users/validate, which is the endpoint that Pushover is now heavily rate limiting.

If chump _doesn’t_ attempt to validate the `device` parameter for you, there’s literally no way to know you got it wrong save for hearing back from an angry message recipient that every single device attached to their account went off, and I can’t in good conscience ship code that defaults to poor user experiences.

If you can’t cache the `User` object, you can set `user.devices = {self.config['device']}` before sending a message. If you _know_ `self.config['device']` is valid there’ll be no issue here and you’ll avoid that endpoint call. You could do the same for `app.sounds` and `self.config[‘sound']` as well if you’d like.

thp:
@karanlyons For what it's worth, `urlwatch` is usually a short-lived process, being run either interactively or (more commonly) through `crontab`, so it's not really that easy to cache objects. Then again, the chump object should only be instantiated / actually send data when there's a website change detected, so it might be "good enough" for what it does (but I don't use pushover, so other users might be better to judge how well it works).

karanlyons:
Your best bet is to then just to set the `user.devices` attribute manually to avoid a request down the wire. It does mean that you won’t know if you got it wrong, but that’s unfortunately the tradeoff Pushover is forcing us to make with their API design as it currently stands. So to avoid _all_ requests beyond sending the message (with the caveat that sending the message may now fail for having a bad sound parameter, _or_ “fail” by succeeding but sending to _all_ devices belonging to the user as opposed to the one you specified):

```python
def web_service_get(self):
    app = chump.Application(self.config['app'])
    if self.config['sound']: app.sounds = {self.config['sound']}
    user = app.get_user(self.config['user'])
    if self.config['device']: user.devices = {self.config['device']}
    return user
```

With `chump>=1.6.0` only one request will then be made. As far as I know Pushover isn’t rate limiting the `sounds` endpoint, so you can omit `if self.config['sound']: app.sounds = {self.config['sound’]}`; 1.6.0 also adds connection pooling for requests, so the overhead for making that additional request should be very minimal. It’s really only their design around devices that’s causing problems.

thp:
@nscaife any feedback to my [comments](
https://github.com/thp/urlwatch/pull/305#pullrequestreview-171246960)?

thp:
Closing for now, please reopen if you have addressed my [comments](
https://github.com/thp/urlwatch/pull/305#pullrequestreview-171246960).

PR #310, opened 2018-11-05
render browser jobs' web pages asynchronously

cfbao:
An alternative of #298. Opening a new pull request due to very significant differences.

Re-implement `BrowserJob` as a subclass of a new class `AsyncJob`, and directly use `pyppeteer` without using `requests_html`.

`worker.py` is still changed, but to a lesser extent, with only additions but no restructuring.

Design of `AsyncJob` workflow:
* Before `run_parallel`: All async jobs are given the same main event loop. The event loop is then put onto an independent thread and starts running.
* During `run_parallel`: No code change in `worker.py` in this stage. In each worker thread, the `retrieve` method of any async job schedules tasks onto the main event loop, and wait for the result and return it.
* After `run_parallel`: The event loop is stopped, and async jobs have a chance to do final clean up.

This is the best method I came up with that can utilize the efficiency of `pyppeteer`'s async API without making urlwatch's code full blown async.
It's also my first time using asyncio / coroutines. Suggestions/criticism more than welcome.

### DIFF
--- a/lib/urlwatch/worker.py
+++ b/lib/urlwatch/worker.py
@@ -30,11 +30,13 @@
 
 import concurrent.futures
 import logging
+import asyncio
+from threading import Thread
 
 import requests
 
 from .handler import JobState
-from .jobs import NotModifiedError
+from .jobs import AsyncJob, NotModifiedError
 
 logger = logging.getLogger(__name__)
 
@@ -48,12 +50,24 @@ def run_parallel(func, items):
         if exception is not None:
             raise exception
         yield future.result()
+    executor.shutdown(wait=False)
 
 
 def run_jobs(urlwatcher):
     cache_storage = urlwatcher.cache_storage
     jobs = urlwatcher.jobs
     report = urlwatcher.report
+
+    async_jobs = [job for job in jobs if isinstance(job, AsyncJob)]
+    loop = asyncio.get_event_loop() if async_jobs else None
+
+    # set up async jobs before the event loop starts running
+    for job in async_jobs:
+        job.setup(loop)
+    # start running the event loop which drives all async jobs
+    if loop:
+        loop_thread = Thread(target=loop.run_forever)
+        loop_thread.start()
 
     logger.debug('Processing %d jobs', len(jobs))
     for job_state in run_parallel(lambda job_state: job_state.process(),
@@ -102,3 +116,12 @@ def run_jobs(urlwatcher):
         else:
             report.new(job_state)
             job_state.save()
+
+    # stop the event loop after all jobs are done
+    if loop:
+        loop.call_soon_threadsafe(loop.stop)
+        loop_thread.join()
+
+    # clean up async jobs after the event loop has stopped
+    for job in async_jobs:
+        job.cleanup()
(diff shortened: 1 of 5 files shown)


cfbao:
A downside I see with this design is that dummy threads are created for every async job. They take up space in the thread pool, but the real work is done (asynchronously) in `loop_thread`. As a result, there are two layers of `concurrent.futures.Futuer` we need to handle.

However, a more efficient implementation would require significant changes in both `worker.py` and `handler.py`.

lib/urlwatch/worker.py
```diff
@@ -48,13 +50,25 @@ def run_parallel(func, items):
...
+    executor.shutdown(wait=False)
 
 
 def run_jobs(urlwatcher):
     cache_storage = urlwatcher.cache_storage
     jobs = urlwatcher.jobs
     report = urlwatcher.report
 
+    async_jobs = [job for job in jobs if isinstance(job, AsyncJob)]
+    loop = asyncio.get_event_loop() if async_jobs else None
```
cfbao: Not sure if I should use `asyncio.new_event_loop` instead.

ftruzzi:
This feature looks awesome but it doesn't wait for dynamic pages to load completely before returning.

It looks like pyppeteer has a `page.waitForSelector` function ([docs](https://miyakogi.github.io/pyppeteer/reference.html#pyppeteer.page.Page.waitForSelector)), it would be nice to have it as a configurable option for every site.

cfbao:
I do have plan to expose more `pyppeteer` features to users. Waiting is certainly one of them. But I'd rather have this "skeleton" code reviewed and approved first, before working on more features.

cfbao:
I'm finding myself disliking this implementation.
If I understand the design correctly, Job classes are meant to be stateless and only store static job descriptions, whereas all state info should go into JobState.
Both `BrowserJob.browser` and `async_job.loop` are state info, and violate this separation. They make job serialization & unserialization unsafe. I'm working on [new features](https://github.com/cfbao/urlwatch/commit/624ee8ea965f30e6c7ee690dc94e5a5a7760835b) where this can actually be problematic.

I'll try to rework the code to put the event loop and browser instance into `JobState`, but not yet quite sure how to do this in a generalizable (e.g. subclass friendly) fashion, especially since both of them are _shared_ external resources.

Would love to hear @thp's input / suggestions if you've got time.

thp:
@cfbao I'm okay with you reworking this feature :)

cfbao:
I have a new implementation in [this branch](https://github.com/cfbao/urlwatch/tree/browser-state).
[Comparison with master](https://github.com/thp/urlwatch/compare/master...cfbao:browser-state).
This version introduces even fewer changes. No `AsyncJob` class. No centrally managed event loop.

Instead, it adds two new methods `request_resources` and `release_resources` to `JobBase`. They are called before and after jobs are run (e.g. the main loop in `worker.py`) respectively.
The resource object (just a dict) is supplied to `JobState.__init__`, and can then be used when processing jobs (e.g. in `job.retrieve`).

Using these two new methods, `BrowserJob` treats the browser instance and the event loop (together with `loop_thread`) as shared resources, and manages them itself.
Future new job classes that use coroutines can create and mange their own independent event loop and "loop thread".

Please take a look and let me know what you think. If you prefer this new one, I can either rebase/overwrite on this pull request, or open a new pull request, whichever you prefer.

thp:
That branch you linked looks good, let's make a PR for that one and start reviewing that and merging it?

cfbao:
Closing in favour of #355.

PR #318, opened 2018-11-10
Make tests run on Windows

cfbao:
* Replace `tempfile.NamedTemporaryFile` with `tempfile.mkstemp`.
The `name` attribute of the file-like object returned by `tempfile.NamedTemporaryFile` cannot be used to open the file again on Windows. See [official docs](https://docs.python.org/3/library/tempfile.html#tempfile.NamedTemporaryFile).

* Skip `UrlsYaml.load_secure` test on Windows.

* Add `cache_storage.close()` in a `finally` block before tear down.
On Windows, opened files cannot be deleted without closing first.

### DIFF
--- a/test/test_handler.py
+++ b/test/test_handler.py
@@ -31,15 +31,20 @@ def test_save_load_jobs():
         ShellJob(name='list proc', command='ls /proc'),
     ]
 
-    with tempfile.NamedTemporaryFile() as tmp:
-        UrlsYaml(tmp.name).save(jobs)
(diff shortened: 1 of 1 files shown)


cfbao:
Admittedly, adding a `finally` block when there's a `teardown` function seems quite unsatisfying. But the `teardown` function doesn't take argument from the test function. Perhaps we can use a test class instead of a function?

thp:
Should be fine for now. I quite like `nose`, but if you want to port it to a newer framework, I'm happy with that as well.

PR #321, opened 2018-11-12
Add CSS selector filter

### LINKED ISSUE #273: Add support for a HTML filter using CSS selectors
Split off from #226, it was suggested that there should be a filter option for using CSS selectors to pick something from the document.

We probably can't use anything from the standard Python library, but something like this should work:
https://cssselect.readthedocs.io/en/latest/

Note that with Requests-HTML (which we already have as an optional dependency for the "browser" job) it would also work: https://html.python-requests.org

cfbao:
Close #273

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,11 +27,12 @@ urlwatch 2 requires:
   * [keyring](https://github.com/jaraco/keyring/)
   * [appdirs](https://github.com/ActiveState/appdirs)
   * [lxml](https://lxml.de)
+  * [cssselect](https://cssselect.readthedocs.io)
   * [enum34](https://pypi.org/project/enum34/) (Python 3.3 only)
 
 The dependencies can be installed with (add `--user` to install to `$HOME`):
 
-`python3 -m pip install pyyaml minidb requests keyring appdirs lxml`
+`python3 -m pip install pyyaml minidb requests keyring appdirs lxml cssselect`
 
 
 Optional dependencies (install via `python3 -m pip install <packagename>`):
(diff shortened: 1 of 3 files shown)


cfbao:
Implement class `LxmlParser` which does all the real work for `CssFilter` and `XPathFilter`.
Similar to the relation and interface between `ElementsBy` and `GetElementBy*`.

README.md
```diff
@@ -150,6 +151,17 @@ filter: xpath:/body
...
+To filter based on a [CSS selector](https://www.w3.org/TR/2011/REC-css3-selectors-20110929/),
+you can use the `css` filter like so:
+
+```yaml
+url: https://example.net/
+filter: css:body
+```
+
+Some limitatons and extensions exist as explained in
+[cssselect's documentatons](https://cssselect.readthedocs.io/en/latest/#supported-selectors).
```
thp: documentatons -> documentation

README.md
```diff
@@ -150,6 +151,17 @@ filter: xpath:/body
...
 
+To filter based on a [CSS selector](https://www.w3.org/TR/2011/REC-css3-selectors-20110929/),
+you can use the `css` filter like so:
+
+```yaml
+url: https://example.net/
+filter: css:body
+```
+
+Some limitatons and extensions exist as explained in
```
thp: limitatons -> limitations

cfbao:
Oops! Fixed.

thp:
Thanks!

PR #325, opened 2018-11-16
Add `ignore_http_error_codes` attribute to url job

### LINKED ISSUE #203: handle 503 requests
It would be a good feature to handle 503 requests. Sometimes pages are under maintenance, triggering a false positive that is perfectly controllable. The cron jobs would still work normally, but every time a 503 request ou a personalised page appears, the job would ignore those pages.

cfbao:
This provides the ability for users to ignore chosen HTTP errors.
Close #203 

I'm not 100% sure about the name of this new attribute. Would like a second opinion.

The error handling logic is moved from `worker.py` to within job classes, to make job class features more self-contained, and to prevent overblown `worker.py` code. This is especially relevant in anticipation of "browser" jobs' upgrade.

### DIFF
--- a/README.md
+++ b/README.md
@@ -161,6 +161,21 @@ url: https://example.com/
 ignore_connection_errors: true
 ```
 
+Similarly, you might want to ignore some (temporary) HTTP errors on the
+server side:
+
+```yaml
+url: https://example.com/
+ignore_http_error_codes: 503
+```
+
+or ignore all HTTP errors if you like:
+
+```yaml
+url: https://example.com/
+ignore_http_error_codes: 4xx, 5xx
+```
+
 For web pages with misconfigured HTTP headers or rare encodings, it may
 be useful to explicitly specify an encoding from Python's
 [Standard Encodings](https://docs.python.org/3/library/codecs.html#standard-encodings).
--- a/README.md
+++ b/README.md
@@ -166,7 +166,7 @@ server side:
 
 ```yaml
 url: https://example.com/
-ignore_http_error_codes: 503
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -161,6 +161,21 @@ url: https://example.com/
 ignore_connection_errors: true
 ```
 
+Similarly, you might want to ignore some (temporary) HTTP errors on the
+server side:
+
+```yaml
+url: https://example.com/
+ignore_http_error_codes: 503
```
thp: Some other temporary error codes that you could include in this example:

408, 429, 500, 502, 503, 504

lib/urlwatch/jobs.py
```diff
@@ -280,6 +286,25 @@ def add_custom_headers(self, headers):
...
+    def ignore_error(self, exception):
+        if isinstance(exception, requests.exceptions.ConnectionError) and self.ignore_connection_errors:
+            return True
+        elif isinstance(exception, requests.exceptions.HTTPError):
+            status_code = exception.response.status_code
+            if isinstance(self.ignore_http_error_codes, int) and self.ignore_http_error_codes == status_code:
+                return True
+            elif isinstance(self.ignore_http_error_codes, str):
+                ignored_codes = [s.strip().lower() for s in self.ignore_http_error_codes.split(',')]
+                if str(status_code) in ignored_codes or '%sxx' % (status_code // 100) in ignored_codes:
```
thp: This is smart, but should we also support `ignore_http_error_codes` being a list of strings/numbers in the YAML? And what if the user explicitly (erroneously) specifies a HTTP status code as a string? (e.g. `"404"`) Should that be handled too?

PR #326, opened 2018-11-26
Add dedupe mode to `--gc-cache`

cfbao:
If a job experienced errors, multiple versions of the same job output data would be saved in cache.
E.g.

| id | guid | timestamp | data | tries | etag |
|----|------|-----------|------|-------|------|
| 1 | ... | ... | old_data | 0 | ...|
| 2 | ... | ... | old_data | 1 | ...|
| 3 | ... | ... | old_data | 2 | ...|
| 4 | ... | ... | old_data | 0 | ...|
| 5 | ... | ... | new_data | 0 | ...|
| 6 | ... | ... | new_data | 1 | ...|
| 7 | ... | ... | new_data | 2 | ...|
| 8 | ... | ... | new_data | 3 | ...|
| 9 | ... | ... | new_data | 0 | ...|

It may be desirable to remove old duplicate entries, preserving only the first entry of each version:

| id | guid | timestamp | data | tries | etag |
|----|------|-----------|------|-------|------|
| 1 | ... | ... | old_data | 0 | ...|
| 5 | ... | ... | new_data | 0 | ...|

This pull request adds a "dedupe" option to `--gc-cache`, which does the operation described above.
The old behavior is named "purge" and still the default.

### DIFF
--- a/lib/urlwatch/command.py
+++ b/lib/urlwatch/command.py
@@ -168,7 +168,9 @@ class UrlwatchCommand:
         if self.urlwatch_config.features:
             sys.exit(self.show_features())
         if self.urlwatch_config.gc_cache:
-            self.urlwatcher.cache_storage.gc([job.get_guid() for job in self.urlwatcher.jobs])
+            self.urlwatcher.cache_storage.gc(
+                [job.get_guid() for job in self.urlwatcher.jobs], self.urlwatch_config.gc_cache)
+            self.urlwatcher.cache_storage.close()
             sys.exit(0)
         if self.urlwatch_config.edit:
             sys.exit(self.urlwatcher.urls_storage.edit(self.urlwatch_config.urls_yaml_example))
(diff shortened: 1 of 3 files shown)


thp:
Good point, actually i think this would require two things:

1. When saving, we should just update "tries" field, not save a new entry
2. Cleaning, as you implemented here -- but probably run by default

cfbao:
I agree it's better to handle this automatically. But I have some questions regarding the design.

* Do we clean the database _every_ time jobs are run? I might be overthinking this, but on larger cache databases, could such cleaning (that basically loads the entire database) ever get a bit too expensive to run as a routine operation?
* If we do clean the database automatically every time, the `dedupe` command line option is probably not needed?
* If duplicates are never created when saving, then routine cleaning shouldn't be needed. A one-time cleaning operation is still needed for older databases.
* We can add to the database a new type of entry that indicates whether/when the database was cleaned. Then we can avoid unnecessary cleaning on newer databases.

thp:
Should we tackle #347 first? This code might be part of the migration code.

cfbao:
Yeah, I agree. Once database is migrated to the proposed new design, dedupe is probably no longer needed.

thp:

PR #328, opened 2018-11-30
Add option to compare new data with multiple history versions

### LINKED ISSUE #22: Only report on persistent changes
I'm finding that I'm getting a lot of false positives. urlwatch detects changes in a webpage in a particular run. In the next run, it detects a complete reversion of these changes. I see this on multiple websites, with different sections of the page. I'm not sure what is causing it, but it might be due to corrupted incoming files.

Perhaps urlwatch could have the option of caching the last two versions when it detects changes, and only report when changes persist for two runs?

### LINKED ISSUE #275: Option to ignore flapping
I watch a page that sometimes flaps between two states, so e.g. it toggles between "A" and "B".
In this special case I don't want to be notified all the time, but only if there is another state.
An option to perseve some history steps and notify only when none of them matches would be great.

cfbao:
Add `compared_versions` [int] flag to Job class.
Report "unchanged" if `new_data` is in the last `compared_versions` versions.
If not, compare `new_data` with the closest match as determined by difflib.

Close #22
Close #275

Again, would love a second opinion on better flag/variable names, especially the flag `compared_versions` and the variable `job_state.history_data` (which is a dict of `data: timestamp` items).

Edit: On second thought, it might not resolve #22 directly. I can rebase/reword if needed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -362,8 +362,9 @@ url: https://example.com/
 compared_versions: 3
 ```
 
-In ths example, changes are only reported if the webpage becomes different from
-the latest three distinct states. Diff is done relative to the closest match.
+In this example, changes are only reported if the webpage becomes different from
+the latest three distinct states. The differences are shown relative to the
+closest match.
 
 
 MIGRATION FROM URLWATCH 1.x
--- a/README.md
+++ b/README.md
@@ -351,6 +351,21 @@ filter:
 ```
 
 
+COMPARE WITH SEVERAL LATEST SNAPSHOTS
+-------------------------------------
+If a webpage frequently changes between several known stable states, it may be
+desirable to have changes reported only if the webpage changes into a new
+unknown state. You can use `compared_versions` to do this.
+
+```yaml
+url: https://example.com/
+compared_versions: 3
+```
+
+In ths example, changes are only reported if the webpage becomes different from
+the latest three distinct states. Diff is done relative to the closest match.
+
+
 MIGRATION FROM URLWATCH 1.x
 ---------------------------
 
--- a/lib/urlwatch/handler.py
+++ b/lib/urlwatch/handler.py
@@ -47,6 +47,7 @@ class JobState(object):
         self.verb = None
         self.old_data = None
         self.new_data = None
+        self.history_data = {}
         self.timestamp = None
         self.exception = None
         self.traceback = None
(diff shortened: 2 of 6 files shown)


README.md
```diff
@@ -351,6 +351,21 @@ filter:
...
+If a webpage frequently changes between several known stable states, it may be
+desirable to have changes reported only if the webpage changes into a new
+unknown state. You can use `compared_versions` to do this.
+
+```yaml
+url: https://example.com/
+compared_versions: 3
+```
+
+In ths example, changes are only reported if the webpage becomes different from
```
thp: Typo: `ths` -> `this`

README.md
```diff
@@ -351,6 +351,21 @@ filter:
...
+the latest three distinct states. Diff is done relative to the closest match.
```
thp: `Diff is done relative to...` -> `The differences are shown relative to...`

lib/urlwatch/handler.py
```diff
@@ -47,16 +47,20 @@ def __init__(self, cache_storage, job):
...
         self.tries = 0
         self.etag = None
 
     def load(self):
-        self.old_data, self.timestamp, self.tries, self.etag = self.cache_storage.load(self.job, self.job.get_guid())
+        guid = self.job.get_guid()
+        self.old_data, self.timestamp, self.tries, self.etag = self.cache_storage.load(self.job, guid)
         if self.tries is None:
             self.tries = 0
+        if self.job.compared_versions and self.job.compared_versions > 1:
```
thp: Does this need `getattr(self.job, 'compared_versions')` or does it work in all cases (don't remember if we set default values for optional job keys)?

lib/urlwatch/storage.py
```diff
@@ -483,6 +483,20 @@ def load(self, job, guid):
 
         return None, None, 0, None
 
+    def get_history_data(self, guid, count=1):
+        history = {}
+        if count < 1:
+            return history
+        for data, timestamp in CacheEntry.query(self.db, CacheEntry.c.data // CacheEntry.c.timestamp,
+                                                order_by=minidb.columns(CacheEntry.c.timestamp.desc, CacheEntry.c.tries.desc),
+                                                where=(CacheEntry.c.guid == guid)
```
thp: Can you instead move the `&` operator to the line starting with `where=`, and remove `W503` from the list of ignored checks?

lib/urlwatch/handler.py
cfbao: This should work. All optional keys are set in `__init__`:
https://github.com/thp/urlwatch/blob/9b289b8f20f492313145cf70246a1166b6bcc52c/lib/urlwatch/jobs.py#L71-L74

lib/urlwatch/storage.py
cfbao: Do you prefer ignoring W504 instead?

W503 and W504 complete opposite of each other, at least one must be ignored. I chose to ignore W503 following [PEP-8's suggestion](https://www.python.org/dev/peps/pep-0008/#should-a-line-break-before-or-after-a-binary-operator).

Let me know if you do prefer the other way around. Just want to make sure you're aware of this conflict and PEP-8 suggestion.

PR #333, opened 2018-12-08
Add option to exclude elements with xpath and css filters

### LINKED ISSUE #329: Deselect filters?
Is there a way to remove or invert something from the filtered element or site? 

e.g. i choose a div in a site but one of the child's from this divs is a dynamic object. like if the filter starts with ! we can invert, deselect or remove it from the previous input? 

```
url: http://example.org/
filter: 
    - element-by-id:something
    - !element-by-id:dynamic_objecs_id
```

```
url: https://example.net/
filter: 
    - xpath://div[@class='content']
    - !xpath://div[@class='dynamic_objecs_class']
```

If if i want everything from a site except one element

```
url: https://example.net/
filter: !xpath://div[@class='dynamic_objecs_class']
```

And there is something else. I'm not really happy with the email notification and how to see changes. 

Is there a way to trigger a simple external script with some parameters like the name-of-urls.yaml name and url? my intention is to pipe all in a text/html/xml file and commit them to a personal git/svn/hg repository. this way i have a nice history and see the changes on different devices in a easy way. 

u would use the name of the urls.yaml as a initial folder name (i want to use this to group them), the name for the filename and the the url in the header of the text/html/xml file for a quickly open the site again. maybe complete entry from the urls.yaml.

cfbao:
Close #329

### DIFF
--- a/lib/urlwatch/filters.py
+++ b/lib/urlwatch/filters.py
@@ -430,9 +430,9 @@ class LxmlParser:
                     parent.text = parent.text + element.tail if parent.text else element.tail
             parent.remove(element)
 
-    @staticmethod
-    def _reevaluate(element):
-        if LxmlParser._orphaned(element):
+    @classmethod
+    def _reevaluate(cls, element):
+        if cls._orphaned(element):
             return None
         if isinstance(element, etree._ElementUnicodeResult):
             parent = element.getparent()
@@ -476,6 +476,7 @@ class LxmlParser:
             root = etree.fromstring(self.data, self.parser)
         if root is None:
             return []
+        excluded_elems = None
         if self.filter_kind == 'css':
             selected_elems = root.cssselect(self.expression)
             excluded_elems = root.cssselect(self.exclude) if self.exclude else None
@@ -485,7 +486,7 @@ class LxmlParser:
         if excluded_elems is not None:
             for el in excluded_elems:
                 self._remove_element(el)
-        return [el for el in map(LxmlParser._reevaluate, selected_elems) if el is not None]
+        return [el for el in map(self._reevaluate, selected_elems) if el is not None]
 
     def get_filtered_data(self):
         return '\n'.join(self._to_string(element) for element in self._get_filtered_elements())
--- a/lib/urlwatch/filters.py
+++ b/lib/urlwatch/filters.py
@@ -414,7 +414,7 @@ class LxmlParser:
         if parent is None:
             # Do not exclude root element
             return
-        if isinstance(element, str):
+        if isinstance(element, etree._ElementUnicodeResult):
             if element.is_tail:
                 parent.tail = None
             elif element.is_text:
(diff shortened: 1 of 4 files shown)


cfbao:
Just realized that this only handles the case when excluded elements are descendants of selected elements.
It doesn't properly handle cases when excluded elements (or their descendants) are directly selected.
Will try to push a fix soon.

thp:
Needs rebasing.

cfbao:
Rebased.

nille02:
@thp something new on this? its a awesome feature that makes some thing much easier.

lib/urlwatch/filters.py
```diff
@@ -406,6 +408,62 @@ def _to_string(self, element):
...
+            previous = element.getprevious()
+            if element.tail is not None:
+                if previous is not None:
+                    previous.tail = previous.tail + element.tail if previous.tail else element.tail
+                else:
+                    parent.text = parent.text + element.tail if parent.text else element.tail
+            parent.remove(element)
+
+    @staticmethod
+    def _reevaluate(element):
```
thp: Since this function uses `LxmlParser._orphaned`, this should be a `@classmethod`:

```
@classmethod
def _reevaluate(cls, element):
    if cls._orphaned(element):
        return None
    ...
```

lib/urlwatch/filters.py
```diff
@@ -419,9 +477,15 @@ def _get_filtered_elements(self):
...
+            selected_elems = root.cssselect(self.expression)
+            excluded_elems = root.cssselect(self.exclude) if self.exclude else None
         elif self.filter_kind == 'xpath':
-            return root.xpath(self.expression)
+            selected_elems = root.xpath(self.expression)
+            excluded_elems = root.xpath(self.exclude) if self.exclude else None
+        if excluded_elems is not None:
+            for el in excluded_elems:
+                self._remove_element(el)
+        return [el for el in map(LxmlParser._reevaluate, selected_elems) if el is not None]
```
thp: `self._reevaluate` instead of `LxmlParser._reevaluate`

lib/urlwatch/filters.py
```diff
@@ -419,9 +477,15 @@ def _get_filtered_elements(self):
...
             return []
         if self.filter_kind == 'css':
-            return root.cssselect(self.expression)
+            selected_elems = root.cssselect(self.expression)
+            excluded_elems = root.cssselect(self.exclude) if self.exclude else None
         elif self.filter_kind == 'xpath':
-            return root.xpath(self.expression)
+            selected_elems = root.xpath(self.expression)
+            excluded_elems = root.xpath(self.exclude) if self.exclude else None
+        if excluded_elems is not None:
```
thp: If `filter_kind` is not `'css'` and not `'xpath'`, `excluded_elems` will never be set? At least introduce a `else:` branch before and set `excluded_elems = None` or set before the line `if self.filter_kind == 'css':`. Thanks :)

nille02:
Any news on this?

PR #337, opened 2018-12-14
Ensure cache database retrieve the latest job state

### LINKED ISSUE #335: Cache database doesn't always load the latest job state
Currently, cache database finds the latest state of a job by `timestamp: int` and `tries: int`. This can be unreliable. If a job experienced errors, and was quickly retried again with success, cache could record something like:

| id | guid | timestamp | data | tries |
|----|------|-----------|------|-------|
| 1 | ... | 1544755012 | ... | 1 |
| 2 | ... | 1544755012 | ... | 2 |
| 3 | ... | 1544755012 | ... | 0 |

Ordering by timestamp then by tries would give the second row as the latest state, whereas the third row is really the latest.

Two possible solutions on my mind:
1. Save `timestamp` as `float` rather than `int`. This should give enough precision in most cases
2. Use `id` to order records. `id` seems to be strictly increasing, but this behavior is not documented in `minidb`, so I'm also a bit hesitant to use it.

In any case, `tries` should not be used for ordering, since it can reset to zero at any point.

This issue is usually not a problem in daily of urlwatch, but is unavoidable in tests, and must be resolved before tests can be improved.

cfbao:
Change `timestamp`'s type to float.
Don't use `tries` to order entries.
Close #335

### DIFF
--- a/lib/urlwatch/storage.py
+++ b/lib/urlwatch/storage.py
@@ -451,7 +451,7 @@ class CacheDirStorage(CacheStorage):
 
 class CacheEntry(minidb.Model):
     guid = str
-    timestamp = int
+    timestamp = float
     data = str
     tries = int
     etag = str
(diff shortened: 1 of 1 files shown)


cfbao:
Closing in favour of an alternative implementation.

PR #355, opened 2019-01-27
Render browser jobs' web pages asynchronously

cfbao:
An alternative of #310.

> This version introduces even fewer changes. No AsyncJob class. No centrally managed event loop. No state info in any job class.
> 
> Instead, it adds two new methods request_resources and release_resources to JobBase. They are called before and after jobs are run (e.g. the main loop in worker.py) respectively.
> The resource object (just a dict) is supplied to JobState.__init__, and can then be used when processing jobs (e.g. in job.retrieve).
> 
> Using these two new methods, BrowserJob treats the browser instance and the event loop (together with loop_thread) as shared resources, and manages them itself.
> Future new job classes that use coroutines can create and mange their own independent event loop and "loop thread".

---

Things that still bug me a bit:

* Maybe not a big deal (for now at least), but using just a dict to store and handle resources feels a bit ad hoc.

* If any `request_resources` call fails, the entire run would fail. Ideally, exceptions should be caught, and only jobs that need such failed resources would fail.
I could just catch all exceptions in `BrowserJob.request_resources` for now, but it's best to have a general mechanism to handle resource exceptions for all future and custom job classes (ones in `hooks.py`).

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,7 +43,7 @@ Optional dependencies (install via `python3 -m pip install <packagename>`):
   * Pushover reporter: [chump](https://github.com/karanlyons/chump/)
   * Pushbullet reporter: [pushbullet.py](https://github.com/randomchars/pushbullet.py)
   * Stdout reporter with color on Windows: [colorama](https://github.com/tartley/colorama)
-  * "browser" job kind: [requests-html](https://html.python-requests.org)
+  * "browser" job kind: [pyppeteer](https://github.com/miyakogi/pyppeteer), Python 3.6 or newer
   * Unit testing: [pycodestyle](http://pycodestyle.pycqa.org/en/latest/)
 
 
@@ -289,7 +289,7 @@ BROWSER
 -------
 
 If the webpage you are trying to watch runs client-side JavaScript to
-render the page, [Requests-HTML](http://html.python-requests.org) can
+render the page, [Pyppeteer](https://github.com/miyakogi/pyppeteer) can
 now be used to render the page in a headless Chromium instance first
 and then use the HTML of the resulting page.
 
--- a/lib/urlwatch/jobs.py
+++ b/lib/urlwatch/jobs.py
@@ -34,6 +34,7 @@ import logging
 import os
 import re
 import subprocess
+import threading
 import requests
 import urlwatch
 from requests.packages.urllib3.exceptions import InsecureRequestWarning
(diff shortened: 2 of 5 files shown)


zevlag:
Are you still working on this feature?  I went to use a `browser` job today to render the page via JS and found it wasn't working.  This current PR seems to have a conflict.

cfbao:
Not right now. This feature is more complex than most others, and I don't have much time.
I do have a local personal branch where this feature is merged with the master branch. I can publish it if people are interested.

README.md
```diff
@@ -43,7 +43,7 @@ Optional dependencies (install via `python3 -m pip install <packagename>`):
   * Pushover reporter: [chump](https://github.com/karanlyons/chump/)
   * Pushbullet reporter: [pushbullet.py](https://github.com/randomchars/pushbullet.py)
   * Stdout reporter with color on Windows: [colorama](https://github.com/tartley/colorama)
-  * "browser" job kind: [requests-html](https://html.python-requests.org)
+  * "browser" job kind: [pyppeteer](https://github.com/miyakogi/pyppeteer), Python 3.6 or newer
```
thp, line 46: Pyppeteer seems to live at https://github.com/pyppeteer/pyppeteer now

lib/urlwatch/jobs.py
```diff
@@ -154,6 +155,26 @@ def get_guid(self):
...
+        ensure success.
+
+        Args:
+            resources: A dict storing external resources.
+        """
+        ...
+
+    def release_resources(self, resources):
+        """Release external resources requested in `request_resources`."""
+        ...
```
thp, line 176: Sounds like maybe a context manager is the right thing to do here?

cfbao: The idea crossed my mind. Can't remember why I decided against it. I'll look into it.

cfbao:
@thp Thanks for the comments!
I'm still not personally satisfied with the overall design of this. I'll try some rethinking.

thp:
I took your code, and updated it in #546 (moved to `urlwatch.browser` module) and updated the documentation.

This also fixes what you mentioned:

> Maybe not a big deal (for now at least), but using just a dict to store and handle resources feels a bit ad hoc.

(there is still some global state in #546, but it's very specific to browser jobs and encapsulated in `BrowserContext`)

PR #360, opened 2019-02-05
New cache database format

### LINKED ISSUE #335: Cache database doesn't always load the latest job state

### LINKED ISSUE #347: Cache database redesign
I've been thinking about tackling the following issues:
* Support job specific check intervals #148, #171
* Find the latest entry of a job accurately #335
* Avoid cache entries duplication #326
* Export site change/version history #53, #170

I believe a redesign of the cache database can help with all these issues.

#### General idea
* Use one table to store snapshots data, where each row corresponds to a distinct snapshot of retrieved data. Rows are immutable.
* Use another table to store job info, where each row corresponds to a different job, together with its last run state info. Rows are mutable.

Currently, the `CacheEntry` table mixes both snapshot info and last run state info. This makes it inflexible to extend and prone to duplication.

#### Tentative details
* Add a new table called `LastRunState`, with columns `guid`, `name`, `location`, `last_checked_time`, `last_snapshot_id`, `tries`, `etag`...
* Deprecate the `tries` and `etag` columns of `the CacheEntry` table.

Then there's the tricky issue of migration... which I'm not sure how best to do it yet.

cfbao:
Close #347, close #326, close #335.

* Reorganize migration code
  - Move urls file checking out of `migrate_urls` and into `Urlwatch.check_directories`.
  - Move migration function calls to the main script.
  - Move storage objects creation inside `Urlwatch.__init__`.
    Creating a cache DB object opens and changes the DB file, so it's important to do migration first.
  - Don't call gc when migrating cache, so multiple urls files setup can migrate safely.
* Implement new cache DB
  - Redesign `CacheStorage`:
    - Reserve `save` for saving new snapshot
    - Add `update` for changing last run job state
    - Use a namedtuple as `load`'s return value for clarity
  - Add new cache DB class:
    - Use seperate tables to store snapshots data and last run job states
    - Disable write methods of old cache formats
  - Migrate old cache DB to new format "in-place"
  - Reserve `JobState.save` for updating job state
  - Add `JobState.save_new_data` for saving new snapshot
  - Simplify job state saving logic thanks to new cache DB design
* Remove socket module from main script
  - default timeout in `socket` no longer relevant after transition to `requests`
* Add and reorganize storage tests


cfbao:
Rebased.
Added and changed some tests.
Edited PR message to close two more related issues.

cfbao:
@thp I haven't looked at this myself for some time as well 😂. Got a real developer job and many things happening in life.

Yeah, I agree it's quite a big PR and not good for review.
I'll try splitting it into small reasonable pieces.

thp:

PR #368, opened 2019-03-22
add __kind__ and subfilter default (None) to custom filters + add example for custom filter usage

robertschulze:


thp:
If you want to have a filter that you enable via `__kind__`, it should be based off of `filters.FilterBase`, the `filters.AutoMatchFilter` / `filters.RegexMatchFilter` filters are designed to be used without `__kind__` (applied directly to whatever URL/pattern matches).

robertschulze:
I have to admit that I don't understand what you mean by "enable via `__kind__`" / "used without `__kind__`".  I just added the filters to the config file intuitively in the same way as any other filter and realized that the functions need some adjustment in order to work. 

Maybe you can provide an example of how you designed the filters to be used (preferably correcting the example in my request), also to prevent others to fall into the same trap?

cfbao:
There are examples of auto match filters in [hooks.py.example](https://github.com/thp/urlwatch/blob/master/share/urlwatch/examples/hooks.py.example).

Auto match filters can only be used with a `hooks.py`. They have no place in a config file.

robertschulze:
Mmh, but using them in urls.yaml yields the errors I saw and tried to fix.

cfbao:
As I said, they have no place in a config file, by which I mean `urls.yaml`.
Once you've written your own auto match filters in `hooks.py`, they're automatically applied to all jobs, and matched based on `MATCH`.

PR #369, opened 2019-03-25
Enable SMTP using a username != from

mborsetti:
Suggest adding the key 'smtpuser' under 'report' 'email' 'smtp' to contain, if present, the SMTP username to be used when logging on to an SMTP server instead of the email 'from' field as currently hardwired. 

This enables the use of urlwatch with SMTP servers that require a username that is different than the 'from' field, such as Amazon's Simple Email Service (https://aws.amazon.com/ses/)

### DIFF
--- a/README.md
+++ b/README.md
@@ -343,7 +343,7 @@ These are the keys you need to configure:
 - `report/email/from`: `you@verified_domain.com` (edit accordingly)
 - `report/email/method`: `smtp`
 - `report/email/smtp/host`: `email-smtp.us-west-2.amazonaws.com` (edit accordingly)
-- `report/email/smtp/smtpuser`: `ABCDEFGHIJ1234567890` (edit accordingly)
+- `report/email/smtp/user`: `ABCDEFGHIJ1234567890` (edit accordingly)
 - `report/email/smtp/keyring`: `true`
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -332,6 +332,27 @@ keychain. To store the password, run: `urlwatch --smtp-login` and enter your
...
+
+Start the configuration editor: `urlwatch --edit-config`
+
+These are the keys you need to configure:
+
+- `report/email/enabled`: `true`
+- `report/email/from`: `you@verified_domain.com` (edit accordingly)
+- `report/email/method`: `smtp`
+- `report/email/smtp/host`: `email-smtp.us-west-2.amazonaws.com` (edit accordingly)
+- `report/email/smtp/smtpuser`: `ABCDEFGHIJ1234567890` (edit accordingly)
```
thp: There's also the `user` key (optional) for `smtp`, why not use that?

mborsetti:
Yes, makes sense.  Made the edits (unfortunately from web there's no reverting).

Got confused because the default config does not have 'user' under 'smtp', which I suggest be added (see lib/urlwatch/storage.py)

thp:
Merged as squashed commit here: fdd4a4f64917ffe3891130c91edd51447225cada -- thanks!

PR #373, opened 2019-04-03
Use TemporaryDirectory with diff_tool to allow file access on Windows

### LINKED ISSUE #371: Permission denied in Windows using diff_tool
I tried to use `diff_tool` using Python on Windows but get permission denied.  Apparently this is a known issue in the [documentation](https://docs.python.org/3/library/tempfile.html#tempfile.NamedTemporaryFile) and a workaround is shown in [Stack Overflow](https://stackoverflow.com/a/9155528/310601).  Is it possible to use `delete=False` option so that this can work on Windows based systems?

cfbao:
Close #371

### DIFF
--- a/lib/urlwatch/reporters.py
+++ b/lib/urlwatch/reporters.py
@@ -35,6 +35,7 @@ import shlex
 import email.utils
 import itertools
 import logging
+import os
 import sys
 import time
 import cgi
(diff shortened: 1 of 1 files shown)


mguinness:
@cfbao Thanks for your PR, I can confirm that `diff_tool` now works under Windows.

Issue #365, opened 2019-03-09
filter: element-by-id


lancelon:
Filter by element ID doesn't seem to work - seems to only work with very select div class and especially not if there are any spaces or hyphens?

Also (wasn't sure whether to create separate issue...) I have two urls in my urls.yaml, one of them "works" (so long as I've found a compatible element) the other url doesn't, even though it's exactly the same format as the first. I get 

`Traceback (most recent call last):
  File "/usr/local/bin/urlwatch", line 111, in <module>
    urlwatch_command.run()
  File "/usr/local/lib/python3.5/dist-packages/urlwatch/command.py", line 293, in run
    self.handle_actions()
  File "/usr/local/lib/python3.5/dist-packages/urlwatch/command.py", line 179, in handle_actions
    sys.exit(self.test_filter())
  File "/usr/local/lib/python3.5/dist-packages/urlwatch/command.py", line 121, in test_filter
    job = job.with_defaults(self.urlwatcher.config_storage.config)
AttributeError: 'NoneType' object has no attribute 'with_defaults'
root@raspbx:~# urlwatch --test-filter 1`

cfbao:
1. Do you have a minimal working example of the `element-by-id` error? It works fine in my tests.
Also, an HTML id attribute should not contain any space.

2. Yes, it's better to create a separate issue.

PR #383, opened 2019-04-15
Fix --test-filter behaviour when job not found

cfbao:
This PR addresses the second part of #365.


cfbao:
Seems like this is already fixed in master? @thp

thp:
Whoops, yes this is already fixed in master. Sorry I must have missed this PR :/

Issue #379, opened 2019-04-13
Feature Request: check individual Job


nille02:
It would be useful to Update a specific job manual. For example `urlwatch --update-job 1`

This is possible a addition to #347

nille02:
Would a PR for this welcome? I looked at the Code and i guess i can implement this.

thp:
Closing, as the PRs have all been closed and no feedback received so far.

PR #386, opened 2019-04-21
add the option to run a single job

nille02:
It runs a single job and is done. 

i think its useful if you want to add a new url but didn't want to check all jobs.

### DIFF
--- a/lib/urlwatch/command.py
+++ b/lib/urlwatch/command.py
@@ -138,10 +138,10 @@ class UrlwatchCommand:
 
     def run_job(self):
         job = self._find_job(self.urlwatch_config.run_job)
-        job = job.with_defaults(self.urlwatcher.config_storage.config)
         if job is None:
             print('Not found: %r' % (self.urlwatch_config.run_job,))
             return 1
+        job = job.with_defaults(self.urlwatcher.config_storage.config)
         self.urlwatcher.jobs = [job]
         self.urlwatcher.run_jobs()
         self.urlwatcher.close()
--- a/lib/urlwatch/command.py
+++ b/lib/urlwatch/command.py
@@ -139,13 +139,13 @@ class UrlwatchCommand:
     def run_job(self):
         job = self._find_job(self.urlwatch_config.run_job)
         if job is None:
-            # Split the Input and Search for Jobs.
+            #Split the Input and Search for Jobs.
             job_items = [self._find_job(item) for item in self.urlwatch_config.run_job.split(',')]
-            # Remove all None Items
+            #Remove all None Items
             job_items = [job for job in job_items if job is not None]
             job_items = [item.with_defaults(self.urlwatcher.config_storage.config) for item in job_items]
 
-        # Check if Job is None or the Joblist is empty
+        #Check if Job is None or the Joblist is empty
         if job is None and len(job_items) == 0:
             print('Not found: %r' % (self.urlwatch_config.run_job,))
             return 1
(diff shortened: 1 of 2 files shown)


nille02:
This is related to #379

lolilolicon:
Why a *single* job? Consider:

1. Allow specifying multiple jobs, with the syntax `--run-jobs 1,3,5`, or a range `--run-jobs 1-3`.
2. Allowing passing `--run-job` multiple times, each one adding to the list of jobs.
3. Allow exclusion, either `--run-jobs -<n...>` or another twin option.

However, since numerical job `indices` are not fixed for a given job, we should allow specifying jobs with a stable identifier, such as `location` à la `--test-filter JOB`, which should also be extended in the same way: opportunity for code reuse.

Considering that, option 2 + a twin option for exclusion is the better solution. Although, some clever heuristics might make the other shorter syntax possible when only numeric indices are given.

nille02:
I agree to the first part, but my python skills are not so well. This implementations was now a low hanging fruit. But i guess i add the ability to add more jobs in one go.

About the second part. The identification is a copy of test-filter. so if you use the url, he find that job like test-filter

lolilolicon:
Yes, your option `--run-job` is exactly like `--delete` and `--test-filter`, and they should (and do) at least share the same routine for figuring out the job list.

The implication of my post really is that all these options should support specifying multiple jobs at the same time, ideally.

However, as I explained above, since a job can not only be identified with a number, but also "location", we may run into ambiguity even with a simple syntax like `1,3,5` for a list of jobs, because "location" could conceivably contain such a substring, or even, with custom jobs, exactly this string. Which was why I said option 2 in my previous post seemed the better solution.

So I guess leave your pull request as is for now; the user can always run `urlwatch` multiple times.

To properly support specifying multiple jobs for all these options, some more thought is needed.

nille02:
The last **makes** pep8 happy again and with this we can run a list of id or urls without run all jobs. i think about a range of job for now.

lib/urlwatch/command.py
```diff
@@ -136,6 +136,28 @@ def test_filter(self):
         # (ignore_cached) and we do not want to store the newly-retrieved data yet (filter testing)
         return 0
 
+    def run_job(self):
+        job = self._find_job(self.urlwatch_config.run_job)
+        if job is None:
+            # Split the Input and Search for Jobs.
+            job_items = [self._find_job(item) for item in self.urlwatch_config.run_job.split(',')]
```
cfbao: URLs may contain commas. It's probably not a good idea to use commas to separate items.

nille02: I can add a util Funktion to Look for escaped commas and/or ignore commas that are in a Quote (or escaped them before the split). That would bei usefuel for many Others parts too.

cfbao: I suggest supporting specifying the option multiple times instead. It'd be much nicer than dealing with messy escapes.

cfbao:
Honestly, I'm not sure supporting running multiple select jobs in one command is really necessary.

If you just want to run a few jobs for one time, it's not much work to type a few commands (for one time only). If you want to run a group of jobs frequently, it's much better to store them in a separate `urls.yaml` and use the `--urls` option.

nille02:
So i revert the patch for multiple jobs right now.

nille02:
Mh, i use it now some times now and to be honest, i liked it to add more as one job at a time. The question is now, just by the job ids (1,2,3,4)/id range (1-10) and don't support the way by url or use the first attempt and ignore that the search for a job can fail. I looked at the --add option and there is no check for it too. just a split by comma. that was a reason by we can add multiple filter items.

thp:
As for the other concerns, argparse supports [`nargs`](https://docs.python.org/3/library/argparse.html#nargs), which you can set to `*` and use it like `--run-jobs a b c` (to avoid the ambiguity with selecting jobs while still allowing multiple jobs to be selected).

thp:
Closing for now. Please reopen in case you want to update this.

Issue #183, opened 2017-10-16
diff(erent) data types


julianuu:
Thanks a lot for your work! I find it really useful.

A wish: I would like to monitor a website which contains a list of links to pdf files and would like to know if the pdfs change. I could imagine that a way to do this currently, is create a filter which finds the links, downloads the pdfs and turns them into text. But thats not really cool and maybe someone would also like to monitor different files which can not be easily turned into text. Do you know a way to make it work?
EDIT: I just saw there is a sha1sum-filter. Ok that would work for detecting _that_ something has changed.

I have not yet understood how the whole program is structured, but my impression is that it contains this workflow:
`get data -> filter -> store data, diff&report`
is this correct? If yes, one way of making my wish work would be to change this into
`get data -> filter -> store data, diff -> report`
and let filter and diff also pass on the information what the data type is that is being handled and then add a differ that can diff pdfs

What do you think? I know maybe it requires too much restructuring to implement this feature, but at the moment I use a script I wrote myself and yours has way more features and is way more stable and easy to use, so for me it feels like a waste of time to put more work into my crappy version when something like this here already exists ;)

ad-m:
Write script which convert pdf to text in stdout and use ``shell`` instead ``url`` job.

thp:
There's now a `pdf2text` filter implememented in the master branch.

Issue #68, opened 2016-04-03
watch binary files


e-dschungel:
It would be nice if urlwatch would be able to watch changes in binary files like images or PDFs by their hashed content.

thp:
You can do this for local files by just piping it into `sha1sum` or a similar tool or even `hexdump` if you want to see which bytes change. For things like PDF you can also pipe it through a tool that converts the PDF to text, so you can even see text changes.

e-dschungel:
Is there a way for remote files, too? This would my use case.

brbsix:
This is probably a bad idea for a number of reasons, but perhaps something like the following would work for your particular use case?

    curl -Is URL | awk '$1 ~ /Last-Modified:/ {sub(".*: ", ""); print}'

thp:
The next version will have two additional filters that will help here:

 * `sha1sum` - Takes the retrieved content and returns the sha1sum as hexdigest
 * `hexdump` - Takes the retrieve content and returns a hexdump-like representation

The hexdump output doesn't include file offsets, as otherwise the diffs will be quite large.

e-dschungel:
@thp Thanks, sha1sum solves my usecase perfectly.

thp:
Note that this will still download the file every time fully when watched. A more efficient solution would be to only do a HEAD request and then use either ETag or Last-Modified as suggested by @brbsix, but we do not have that feature (getting HTTP headers) yet. We do already support specifying the HTTP method, so maybe if it's HEAD, the content should be the raw headers, since the body will always be empty, anyway.

PR #390, opened 2019-04-27
Cumulative changes: 'pdf2text' filter, fix 'format-json' to handle utf-8, add viewport to HTML

mborsetti:
A few fixes and the introduction of a filter to transform PDF files to text (see https://github.com/thp/urlwatch/issues/68 and https://github.com/thp/urlwatch/issues/183).  Fixed an issue with 'format-json' to handle utf-8 and added viewport to HTML


cfbao:
* I believe it's best practice to limit a PR to a single issue rather than a cumulative update.

* We should probably design a more general mechanism to handle `str` or `bytes` output/input. Deciding by finding one particular filter in the filter list is rather ad hoc. Other filters (e.g. `sha1sum`, `hexdump` and potential custom filters in `hooks.py`) may also use a binary input, and new job types (including custom ones) shouldn't need to worry about every filter's expectation.

* `format-json` already handles Unicode characters correctly. It's simply that it escapes them rather than keeping them as is. It's probably best to leave `ensure_ascii` as a user adjustable option and default to old behaviour to ensure backwards compatibility.

thp:
(btw, those are all great changes that I would like to see in urlwatch, so if you don't want to split them up into separate PRs, let me know, I can also cherry-pick changes manually for you)

mborsetti:
> (btw, those are all great changes that I would like to see in urlwatch, so if you don't want to split them up into separate PRs, let me know, I can also cherry-pick changes manually for you)

**@thp**,

Thanks.  I am new to git/github and this is a nights & weekends project -- you've motivated me to work on learning more about using git/github and resubmit. Most likely not before the weekend (US time).

thp:
They way I'd do it is probably:

For each "group" of changes, `git checkout master`, then `git checkout -b name-of-feature` then `git cherry-pick COMMIT1 COMMIT2 ...` (where `COMMIT1` and `COMMIT2`, ... are the commit IDs of the commits you want to put into that branch) and then push + make a PR.

mborsetti:
I finally had enough free time to learn enough git to be dangerous! :)

Separate pull requests have been placed.

**THANKS** so much for the tips; I would have never found `git cherry-pick` without your hint, and I am so glad I did.

PR #402, opened 2019-05-17
Add the Option to limit the E-Mail subject length

nille02:
The E-Mail Subject is currently not limited. In Clients like Thunderbird at a certain length its no longer possible to read anything else than the subject.

### DIFF
--- a/lib/urlwatch/reporters.py
+++ b/lib/urlwatch/reporters.py
@@ -383,6 +383,8 @@ class EMailReporter(TextReporter):
 
     def submit(self):
         filtered_job_states = list(self.report.get_filtered_job_states(self.job_states))
+        cfg = self.report.config['report']['email']
+        subject_length = cfg['subject_length']
 
         subject_args = {
             'count': len(filtered_job_states),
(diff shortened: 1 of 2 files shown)


lib/urlwatch/reporters.py
```diff
@@ -383,6 +383,8 @@ class EMailReporter(TextReporter):
 
     def submit(self):
         filtered_job_states = list(self.report.get_filtered_job_states(self.job_states))
+        cfg = self.report.config['report']['email']
+        subject_length = cfg['subject_length']
```
thp: This needs to deal with situations where `subject_length` is not set (e.g. old configuration where this key didn't exist). Also, it's probably useful to make it so that there is an option of setting the value to `null` or something so that the subject length isn't limited.

lib/urlwatch/reporters.py
```diff
@@ -383,6 +383,8 @@ class EMailReporter(TextReporter):
 
     def submit(self):
         filtered_job_states = list(self.report.get_filtered_job_states(self.job_states))
+        cfg = self.report.config['report']['email']
```
thp: You should be able to use `self.config` instead of `self.report.config['report']['email']` -- this is also more flexible, as it doesn't hardcode the reporter kind and/or path to the config.

lib/urlwatch/reporters.py
```diff
@@ -407,9 +409,10 @@ def submit(self):
         if self.config['html']:
             body_html = '\n'.join(self.convert(HtmlReporter).submit())
 
-            msg = mailer.msg_html(self.config['from'], self.config['to'], subject, body_text, body_html)
+            msg = mailer.msg_html(self.config['from'], self.config['to'], subject[:subject_length],
```
thp: You cut off the subject in two places now, why not do it above in a single place where the `subject` is first assigned?

thp:
@nille02 did you have time to look at the review comments?

thp:
@nille02 Any news on this? If not, I'm probably going to close it the next time I'm looking at this PR :)

PR #404, opened 2019-05-24
Add XML namespace support to xpath & css filters

### LINKED ISSUE #403: support XML namespace in XPath filter
Trying the following configuration:
```yaml
filter: 
  - xpath:   
      path: '//EntityDescriptor'
      method: xml
kind: url
url: http://www.aai.dfn.de/fileadmin/metadata/dfn-aai-test-metadata.xml
```
I get an empty result instead of all `EntityDescriptor` tags. 

Is this some kind of lxml limitation I am hitting here?

cfbao:
Close #403

### DIFF
--- a/README.md
+++ b/README.md
@@ -421,6 +421,30 @@ filter:
   - html2text: re
 ```
 
+To match an element in an [XML namespace](https://www.w3.org/TR/xml-names/),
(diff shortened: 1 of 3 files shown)


thp:
Thanks!

PR #421, opened 2019-10-08
Matrix reporter

reivilibre:
This adds a reporter compatible with the Matrix.org protocol, as used in e.g. Riot.im.

### DIFF
--- a/lib/urlwatch/reporters.py
+++ b/lib/urlwatch/reporters.py
@@ -57,8 +57,18 @@ try:
 except ImportError:
     Pushbullet = None
 
-logger = logging.getLogger(__name__)
+try:
+    import matrix_client.api
+except ImportError:
+    matrix_client = None
 
+try:
+    # markdown2 is an optional dependency which provides better formatting for Matrix.
+    from markdown2 import Markdown
+except ImportError:
+    Markdown = None
+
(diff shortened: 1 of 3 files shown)


lib/urlwatch/reporters.py
```diff
@@ -57,8 +57,18 @@
...
 
-logger = logging.getLogger(__name__)
+try:
+    import matrix_client.api
+except ImportError:
+    matrix_client = None
+
+try:
+    # markdown2 is an optional dependency which provides better formatting for Matrix.
+    from markdown2 import Markdown
```
thp: This optional dependency should be mentioned in the `README.md` file.

lib/urlwatch/reporters.py
```diff
@@ -614,3 +624,110 @@ def submit_to_slack(self, webhook_url, text):
 
     def chunkstring(self, string, length):
         return (string[0 + i:length + i] for i in range(0, len(string), length))
+
+
+class MarkdownReporter(ReporterBase):
```
thp: Can the `MarkdownReporter` share some/more code with the `TextReporter`? Seems like there's some code duplication there now.

thp:
Closing in favor of #445.

PR #424, opened 2019-10-30
Replace `cgi.escape` with `html.escape`

cfbao:
`cgi.escape` is unsafe and removed in Python 3.8.
`html.escape` is the recommended replacement.


thp:
Thanks!

PR #428, opened 2019-11-15
Add simple replace filter example

robertschulze:
The example is not only instructive but actually even useful, e.g. to insert linebreaks if they are missing in the HTML.

### DIFF
--- a/share/urlwatch/examples/hooks.py.example
+++ b/share/urlwatch/examples/hooks.py.example
@@ -115,3 +115,10 @@ class CustomHtmlFileReporter(reporters.HtmlReporter):
     def submit(self):
         with open(self.config['filename'], 'w') as fp:
             fp.write('\n'.join(super().submit()))
+
+
+class SimpleReplaceFilter(filters.FilterBase):
+    __kind__ = 'simple-replace-filter'
+    def filter(self, data, subfilter=None):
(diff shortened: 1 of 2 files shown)


thp:
This already exists as the `RegexSub` filter (aka `re.sub`), the `README.md` even has an example converting single HTML pieces to plain text (plus it supports leaving out the replacement string, defaulting to the empty string):

```yaml
kind: url
url: https://example.com/
filter:
    - re.sub: '\s*href="[^"]*"'
    - re.sub:
        pattern: '<h1>'
        repl: 'HEADING 1: '
    - re.sub:
        pattern: '</([^>]*)>'
        repl: '<END OF TAG \1>'
```

thp:
By the way, if you want to convert `<br>` to `\n`, maybe you also want to take a look at the existing `html2text` filter? (it also converts all other tags, but it might be what you are looking for). If you just want to convert `<br>` to `\n`, but leave all other HTML tags intact, you do need it, of course, and you can do it like that:
```
url: https://example.com/
filter:
    - re.sub:
        pattern: '<br[/]?>'
        repl: '\n'
```

Note that the use of `<br[/]?>` should match both `<br>` and `<br/>` (it however won't match `<br />` -- this is left as an exercise for the reader..).

PR #431, opened 2019-12-06
Add support for insecure SMTP password storage (Fixes #277)

### LINKED ISSUE #277: always prompt for password by python keyring
I updated to the last version and now each time I launch urlwatch, a password is prompt.
If I put the password it works, but as it is used in a crontab, how to deal without password ?

thp:
As requested in #277, here's a possible way to supply the password as plain text in an insecure way to avoid having to deal with keyring access in cron.


scottmac:
Any chance this can be merged soon?

thp:
@scottmac merged

PR #433, opened 2019-12-07
format-json filter now handles utf-8 correctly

mborsetti:
The world is moving to UTF-8 (see https://discuss.python.org/t/pep-597-use-utf-8-for-default-text-file-encoding/1819).

The current filter outputs non-ASCII characters as escaped (e.g. '–' is replaced by '\u2013') which makes some text very hard to read.  json.dumps forces a transformation of non-ASCII characters, which is no longer needed on modern computers.

Documentation: https://docs.python.org/3/library/json.html#json.dump


thp:
Thanks!

PR #434, opened 2019-12-07
New filter 'pdf2text' to extract text from PDF

mborsetti:
A filter 'pdf2text' is created to extract text from a PDF file; this allows tracking of changes to the contents of PDF files.

It is based on the pdftotext library (https://github.com/jalan/pdftotext), which in turn is based on poppler (https://poppler.freedesktop.org/).  README has been updated with the additional requirements.

It only works on bytes, so I also had to modify code in the job.py library: if this filter is specified, a bytes object is returned instead of a str.  Longer term this probably should be a more generic option.

### DIFF
--- a/docs/source/dependencies.rst
+++ b/docs/source/dependencies.rst
@@ -38,7 +38,7 @@ Where ``<packagename>`` is one of the following:
 +-------------------------+---------------------------------------------------------------------+
 | Feature                 | Python package(s) to install                                        |
 +=========================+=====================================================================+
-| Pushover reporter       | `chump <https://github.com/karanlyons/chump/>`__                    |
+| Pushover reporter       | `chump <https://github.com/karanlyons/chump>`__                     |
 +-------------------------+---------------------------------------------------------------------+
 | Pushbullet reporter     | `pushbullet.py <https://github.com/randomchars/pushbullet.py>`__    |
 +-------------------------+---------------------------------------------------------------------+
@@ -55,5 +55,8 @@ Where ``<packagename>`` is one of the following:
 | `beautify` filter       | `beautifulsoup4 <https://pypi.org/project/beautifulsoup4/>`__,      |
 |                         | `jsbeautifier <https://pypi.org/project/jsbeautifier/>`__ and       |
 |                         | `cssbeautifier <https://pypi.org/project/cssbeautifier/>`__         |
++-------------------------+---------------------------------------------------------------------+
+| `pdf2text` filter       | `pdftotext <https://github.com/jalan/pdftotext>`__ and              |
+|                         | its OS-specific dependencies (see the above link)                   |
 +-------------------------+---------------------------------------------------------------------+
 
--- a/lib/urlwatch/filters.py
+++ b/lib/urlwatch/filters.py
@@ -206,16 +206,17 @@ class Pdf2TextFilter(FilterBase):
     # Note: check pdftotext website for OS-specific dependencies for install
 
     __kind__ = 'pdf2text'
+    __uses_bytes__ = True
 
     def filter(self, data, subfilter=None):
         # data must be bytes
+        if not isinstance(data, bytes):
+            raise ValueError('The pdf2text filter needs bytes input (is it the first filter?)')
 
         if subfilter is None:
             password = ''
         elif isinstance(subfilter, dict):
             password = subfilter['password']
-        elif isinstance(subfilter, str):
-            password = subfilter
         import pdftotext
         import io
         return '\n\n'.join(pdftotext.PDF(io.BytesIO(data), password=password))
--- a/lib/urlwatch/filters.py
+++ b/lib/urlwatch/filters.py
@@ -198,6 +198,29 @@ class Html2TextFilter(FilterBase):
         return html2text(data, method=method, options=options)
 
 
+class Pdf2TextFilter(FilterBase):
+    """Convert PDF to plaintext"""
+    # Requires data to be in bytes (not unicode)
+    # Dependency: pdftotext (https://github.com/jalan/pdftotext), itself based
+    # on poppler (https://poppler.freedesktop.org/)
+    # Note: check pdftotext website for OS-specific dependencies for install
+
+    __kind__ = 'pdf2text'
+
+    def filter(self, data, subfilter=None):
+        # data must be bytes
+
+        if subfilter is None:
+            password = ''
+        elif isinstance(subfilter, dict):
+            password = subfilter['password']
+        elif isinstance(subfilter, str):
+            password = subfilter
+        import pdftotext
+        import io
+        return '\n\n'.join(pdftotext.PDF(io.BytesIO(data), password=password))
+
+
 class Ical2TextFilter(FilterBase):
     """Convert iCalendar to plaintext"""
 
--- a/lib/urlwatch/filters.py
+++ b/lib/urlwatch/filters.py
@@ -89,6 +89,11 @@ class FilterBase(object, metaclass=TrackSubClasses):
             raise ValueError('Unknown filter kind: %s:%s' % (filter_kind, subfilter))
         return filtercls(state.job, state).filter(data, subfilter)
 
+    @classmethod
+    def is_bytes_filter(cls, filter):
+        return (filter in [name for name, class_ in FilterBase.__subclasses__.items()
+                           if getattr(class_, '__uses_bytes__', False)])
+
     def match(self):
         return False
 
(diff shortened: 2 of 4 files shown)


cfbao:
You've merged your master branch (which contains all changes you want to make) into this single feature branch, which resulted in this PR containing all your proposed changes (instead of a single one).

Your commit [d46b4c9](https://github.com/mborsetti/urlwatch/tree/d46b4c90cccb3911e508e93ec8015335eb379d57) seems to be a good place for this branch and PR.
You can run the following git commands to set this branch to this commit:
```
git checkout pdf2text
git reset --hard d46b4c9
git push --force
```

cfbao:
My first thought on jobs/filters' input/output types (str/bytes) handling:

A filter class can declare its supported input and output types as class variables
```python
class ExampleFilter(FilterBase):
    INPUT_TYPE = str
    OUTPUT_TYPE = [str, bytes]
```
The `Filter.filter` method and `job.retrieve` method should have an optional argument `output_type` that's used when multiple output types are supported.

The logic of matching and using correct input/output types should be placed outside of jobs or filters, but inside `JobState.process`.

The base filter class should declare default input/output types both as `str` for backwards compatibility, to not break users' custom hooks.

mborsetti:
> You've merged your master branch (which contains all changes you want to make) into this single feature branch, which resulted in this PR containing all your proposed changes (instead of a single one).
> 
> Your commit [d46b4c9](https://github.com/mborsetti/urlwatch/tree/d46b4c90cccb3911e508e93ec8015335eb379d57) seems to be a good place for this branch and PR.
> You can run the following git commands to set this branch to this commit:
> 
> ```
> git checkout pdf2text
> git reset --hard d46b4c9
> git push --force
> ```

Thanks, and apologies for the inadvertent mess (I still have to understand how I did it, I think it was on the website).

lib/urlwatch/jobs.py
```diff
@@ -284,6 +287,11 @@ def retrieve(self, job_state):
         # Save ETag from response into job_state, which will be saved in cache
         job_state.etag = response.headers.get('ETag')
 
+        # If 'pdf2text' filter is selected, return bytes content as opposed to str
+        # as it's required by the library used by that filter
+        if 'pdf2text' in self.filter or any('pdf2text' in subfilter for subfilter in self.filter):
```
thp: Can you turn this expression (`'pdf2text' in self.filter or any('pdf2text' in subfilter for subfilter in self.filter)`) into a helper function like `self.needs_binary_data()`? Or maybe we could make it so that this job has a "binary: true" flag or something? IMHO it would be better to do it this way (or make it so that the filter object has a "needs_binary = True" or something and find it out this way -- right now it's a bit hard to maintain).

thp:
@mborsetti Did you have time to look at the PR review comments already?

mborsetti:
@thp,

On my todo list now; should be done tonight UTC +08.

mborsetti:
I am sorry, but due to some developments relating to COVID-19 I have not
been able to do any and highly doubt I will be able to contribute at all in
the near term.

mborsetti:
Hi Tom,
Had a bit of time to look at this. I am still a bit confused about the overall architecture. If a switch indicating "filter needs binary data" or "job is binary data" were to be created, where would it be set (i.e. where would you do the conditional `if 'pdf2text' in self.filter or any('pdf2text' in subfilter for subfilter in self.filter):`?

lib/urlwatch/filters.py
```diff
@@ -198,6 +198,29 @@ def filter(self, data, subfilter=None):
...
+
+    def filter(self, data, subfilter=None):
+        # data must be bytes
+
+        if subfilter is None:
+            password = ''
+        elif isinstance(subfilter, dict):
+            password = subfilter['password']
+        elif isinstance(subfilter, str):
+            password = subfilter
```
thp, line 218: IMHO it's totally fine to require that the subfilter must be specified as dict, so this can be removed.

docs/source/dependencies.rst
```diff
@@ -38,7 +38,7 @@ Where ``<packagename>`` is one of the following:
 +-------------------------+---------------------------------------------------------------------+
 | Feature                 | Python package(s) to install                                        |
 +=========================+=====================================================================+
-| Pushover reporter       | `chump <https://github.com/karanlyons/chump/>`__                    |
+| Pushover reporter       | `chump <https://github.com/karanlyons/chump>`__                     |
```
thp, line 41: Unrelated change to this PR, can you move it into a separate PR (yeah, it's a one-character change, but just to keep it separate).

mborsetti: Fair enough!

lib/urlwatch/filters.py
mborsetti: 'k

lib/urlwatch/jobs.py
```diff
@@ -284,6 +284,11 @@ def retrieve(self, job_state):
         # Save ETag from response into job_state, which will be saved in cache
         job_state.etag = response.headers.get('ETag')
 
+        # If 'pdf2text' filter is selected, return content in bytes instead of in unicode
+        # as that's what's required by the library used by that filter
+        if self.filter and ('pdf2text' in self.filter or any('pdf2text' in subfilter for subfilter in self.filter)):
+            return response.content
+
```
thp, line 291: "subfilter" is what we call the "parameters" of the filter. Here's my proposal:

- Add `__uses_bytes__ = True` to `Pdf2TextFilter`

Then, to get all filter names:

    bytes_filters = [name for name, class_ in FilterBase.__subclasses__.items()
                     if getattr(class_, '__uses_bytes__', False)]

(by default this should evaluate to `['pdf2text']` if `__uses__bytes__ = True` is set on the `Pdf2TextFilter` class)

Then you can do:

    for filter_name in bytes_filters:
        if self.filter and (filter_name in self.filter or any(filter_name in filter for filter in self.filter)):
            # a bytes filter is selected, return content in bytes instead of in unicode
            # as that's what's required by the library used by that filter
            return response.content

Of course, it's tricky if you try to chain filters... This assumes that the `pdf2text` filter is the "first" filter. The following filter chain won't work (properly):

```
    filter:
        - html2text
        - pdf2text
```

So maybe even check if it's the only filter or if `isinstance(self.filter, list)` then only check if the first filter is a bytes filter (and maybe also check that all others filters are NOT bytes filters).

mborsetti: Checking the first filter makes more sense since Job only creates data to the first filter

The issue with chaining only makes sense if you have filters that output bytes, which you don't have.  Also, I can't think what would be a use case for one of them.  Will document that pdf2text needs to be the *first* filter and work on throwing an error if it's fed a str type, which would happen if it's not the first filter.

thp:
Thanks, will merge once CI is happy.

lib/urlwatch/jobs.py
```diff
@@ -198,6 +199,13 @@ def retrieve(self, job_state):
         if result != 0:
             raise ShellError(result)
 
+    bytes_filters = [name for name, class_ in FilterBase.__subclasses__.items()
+                     if getattr(class_, '__uses_bytes__', False)]
```
thp, line 203: Now that is in two places, maybe we can make this a classmethod on `FilterBase`?

    @classmethod
    def is_bytes_filter(cls, filter):
        return (filter in [name for name, class_ in FilterBase.__subclasses__.items()
                           if getattr(class_, '__uses_bytes__', False)])

mborsetti: Oh, I am learning lots from you.  On to it.

lib/urlwatch/jobs.py
```diff
@@ -198,6 +199,11 @@ def retrieve(self, job_state):
         if result != 0:
             raise ShellError(result)
 
+        # If the first filter is a bytes filter, return content in bytes instead of in unicode
+        # as that's what's required by the library used by that filter
+        if self.filter and (FilterBase(self, self.filter) or FilterBase(self, self.filter[0])):
```
thp, line 204:
```python
if self.filter and (Filterbase.is_bytes_filter(self.filter) or FilterBase.is_bytes_filter(self.filter[0]):
```

mborsetti: Yeah, my internal test needs upgrading; thanks for the hint.

The correct code is actually `if self.filter and (FilterBase.is_bytes_filter(self.filter)
                            or any(FilterBase.is_bytes_filter(filter) for filter in self.filter[0])):`, although it's still technically incorrect because I should be checking the first key only, not all (using `any()`)

mborsetti: Now only checking first key.

I can't think of anything else.  Thanks for your guidance and patience.

thp:
Thanks for working on this! Merged :)

PR #443, opened 2019-12-28
Allow multiple email addresses for the SMTP option

scottmac:
SMTP.sendmail() can take a list of email addresses. This change is backwards compatible so if you don't have a comma everything still works.


scottmac:
Comma separated addresses already works with sendmail it's their default, that's why I added it just to the SMTP class to make it compatible.

scottmac:
I'll work on a change to make this more YAML friendly too

PR #444, opened 2020-01-04
Correct issue with terminal url selector picking up ending bracket

caltlgin:
When displaying urls in certain terminals, the ending bracket of the url string place holder is incorporated into the final url.

**eg:** https://github.com/thp/urlwatch)

This merge has put a single space at the beginning and end of the url place holder, so now the end bracket will no longer become part of the url.


thp:
Thanks!

PR #445, opened 2020-02-02
Matrix reporter

schildbach:
I took the top three commits from https://github.com/reivilibre/fork-urlwatch-matrix/commits/master, squashed them and rebased on the last urlwatch release. Then I successfully verified that reporting via Matrix works.

I added my own corrections as separate commit(s). See the respective commit messages.

One problem I'm seeing is that the [matrix-python-sdk](https://github.com/matrix-org/matrix-python-sdk) dependency is not maintained any more (see their readme).

Note: I'm not a Python developer, so I only marginally reviewed the actual code.

### DIFF
--- a/lib/urlwatch/reporters.py
+++ b/lib/urlwatch/reporters.py
@@ -58,8 +58,18 @@ try:
 except ImportError:
     Pushbullet = None
 
-logger = logging.getLogger(__name__)
+try:
+    import matrix_client.api
+except ImportError:
+    matrix_client = None
 
+try:
+    # markdown2 is an optional dependency which provides better formatting for Matrix.
+    from markdown2 import Markdown
+except ImportError:
+    Markdown = None
+
+logger = logging.getLogger(__name__)
 
 # Regular expressions that match the added/removed markers of GNU wdiff output
 WDIFF_ADDED_RE = r'[{][+].*?[+][}]'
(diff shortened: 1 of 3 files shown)


schildbach:
I forgot to notify @reivilibre that his/her code is in this PR.

schildbach:
We need a max message length check. If messages exceed 65k bytes/chars, we get:

matrix_client.errors.MatrixRequestError: 413: {"errcode":"M_TOO_LARGE","error":"event too large"}

thp:
@schildbach the check is already in there as `MAX_LENGTH = 4096`, right?

schildbach:
> @schildbach the check is already in there as `MAX_LENGTH = 4096`, right?

It doesn't seem to be effective. As described in my last comment, I've had a large message fail.

I also see no reference to the `MAX_LENGTH` variable. The other reporters have a code block like this:

```
        if len(body_text) > self.MAX_LENGTH:
            body_text = body_text[:self.MAX_LENGTH]
```

schildbach:
I've just pushed a commit on top that adds the missing body trim.

lib/urlwatch/reporters.py
```diff
@@ -625,3 +635,113 @@ def submit_to_slack(self, webhook_url, text):
...
+        room_id = self.config['room_id']
+
+        body_markdown = '\n'.join(super().submit())
+
+        if not body_markdown:
+            logger.debug('Not calling Matrix API (no changes)')
+            return
+
+        if len(body_markdown) > self.MAX_LENGTH:
+            body_markdown = body_markdown[:self.MAX_LENGTH]
```
thp: Maybe add an ellipsis or something (be sure to subtract the additional characters from `MAX_LENGTH` so that the resulting string is still below `MAX_LENGTH` characters) so that it's obvious that some content was cut off, otherwise if somebody is super unlucky it will be cut off and they won't know that it was cut off and miss a notification.

thp:
@schildbach Had time to look at my comment? Other than that, I think the PR is ready to be merged.

lib/urlwatch/reporters.py
schildbach: I simply copied the code from other reporters. The fix to the problem you've described should probably be applied to all such occurences uniformly.

thp:
Merged, thanks @schildbach and @reivilibre :)

PR #455, opened 2020-02-25
Use pypa recommended method for supplying scripts

### LINKED ISSUE #451: Windows: add console_scripts entry point?
First time using urlwatch, unfortunately have to use it in a Windows environment.

I installed urlwatch using [pipx](https://github.com/pipxproject/pipx), and it dropped the urlwatch script from the repo parent directory into my local bin dir, however it is not runnable as a command on Windows.

Other python packages, for example `pycowsay`, define a [console entry point](https://github.com/cs01/pycowsay/blob/b60f4f3b688592ea0924dcf5820b0b094fbf2c95/setup.py#L48) which allows for an [executable to be created](https://packaging.python.org/specifications/entry-points/#use-for-scripts) on install. 

I tried adding an entry point to urlwatch's setup.py but it seems like the package structure doesn't allow for an entry point to be created since the urlwatch script is in the same directory as setup.py.

How are other people on Windows using urlwatch?
(For the time being I have installed it in a venv and am manually calling python /path/to/urlwatch)

ammgws:
See https://packaging.python.org/guides/distributing-packages-using-setuptools/#scripts

> Although setup() supports a scripts keyword for pointing to pre-made scripts to install, the recommended approach to achieve cross-platform compatibility is to use console_scripts entry points (see below).

Fixes #451

### DIFF
--- a/lib/urlwatch/cli.py
+++ b/lib/urlwatch/cli.py
@@ -84,7 +84,7 @@ def setup_logger(verbose):
         root_logger.info('turning on verbose logging mode')
 
 
-if __name__ == '__main__':
+def main():
     config_file = os.path.join(urlwatch_dir, CONFIG_FILE)
     urls_file = os.path.join(urlwatch_dir, URLS_FILE)
     hooks_file = os.path.join(urlwatch_dir, HOOKS_FILE)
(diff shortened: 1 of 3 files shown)


thp:
Thanks!

PR #457, opened 2020-02-27
Change slack to 40000 bytes

scottmac:
Slack currently supports 40,000 characters, the 4096 was a suggested length that was much lower than supported.

https://api.slack.com/changelog/2018-04-truncating-really-long-messages


thp:
Thanks!

PR #459, opened 2020-03-01
Provide hooks with content_type (raw from server) and mime_type

tomaszn:
This allows creating automatic converters like:
```
class AtomFeedFilter(AutoMatchFilter):           
    MATCH = {'mime_type': 'application/atom+xml'}
    def filter(self, data):
        ...
```

### DIFF
--- a/lib/urlwatch/jobs.py
+++ b/lib/urlwatch/jobs.py
@@ -289,7 +289,9 @@ class UrlJob(Job):
         # urlwatch behavior and try UTF-8 decoding first.
         content_type = response.headers.get('Content-type', '')
         content_type_match = self.CHARSET_RE.match(content_type)
-        if not content_type_match and not self.encoding:
+        result = None
+
+        def try_decoding(response):
             try:
                 try:
                     try:
@@ -301,10 +303,14 @@ class UrlJob(Job):
             except LookupError:
                 # If this is an invalid encoding, decode as ascii (Debian bug 731931)
                 return response.content.decode('ascii', 'ignore')
+
+        if not content_type_match and not self.encoding:
+            result = try_decoding(response)
+        else:
+            result = response.text
         if self.encoding:
             response.encoding = self.encoding
-
-        return response.text
+        return result
 
     def add_custom_headers(self, headers):
         """
--- a/lib/urlwatch/jobs.py
+++ b/lib/urlwatch/jobs.py
@@ -209,7 +209,7 @@ class UrlJob(Job):
     __required__ = ('url',)
     __optional__ = ('cookies', 'data', 'method', 'ssl_no_verify', 'ignore_cached', 'http_proxy', 'https_proxy',
                     'headers', 'ignore_connection_errors', 'ignore_http_error_codes', 'encoding', 'timeout',
-                    'ignore_timeout_errors', 'ignore_too_many_redirects')
+                    'ignore_timeout_errors', 'ignore_too_many_redirects', 'content_type', 'mime_type')
 
     LOCATION_IS_URL = True
     CHARSET_RE = re.compile('text/(html|plain); charset=([^;]*)')
--- a/lib/urlwatch/jobs.py
+++ b/lib/urlwatch/jobs.py
@@ -198,7 +198,7 @@ class ShellJob(Job):
         if result != 0:
             raise ShellError(result)
 
-        return stdout_data.decode('utf-8')
+        return (stdout_data.decode('utf-8'), 'text/plain')
 
 
 class UrlJob(Job):
@@ -254,7 +254,7 @@ class UrlJob(Job):
         file_scheme = 'file://'
         if self.url.startswith(file_scheme):
             logger.info('Using local filesystem (%s URI scheme)', file_scheme)
-            return open(self.url[len(file_scheme):], 'rt').read()
+            return (open(self.url[len(file_scheme):], 'rt').read(), 'text/plain')
 
         if self.headers:
             self.add_custom_headers(headers)
@@ -310,7 +310,7 @@ class UrlJob(Job):
             result = response.text
         if self.encoding:
             response.encoding = self.encoding
-        return result
+        return (result, content_type)
 
     def add_custom_headers(self, headers):
         """
@@ -364,4 +364,4 @@ class BrowserJob(Job):
         from requests_html import HTMLSession
         session = HTMLSession()
         response = session.get(self.navigate)
-        return response.html.html
+        return (response.html.html, response.headers.get('Content-type', ''))
--- a/share/urlwatch/examples/hooks.py.example
+++ b/share/urlwatch/examples/hooks.py.example
@@ -41,7 +41,10 @@ from urlwatch import reporters
 #    __required__ = ('username', 'password')
 #
 #    def retrieve(self, job_state):
-#        return 'Would log in to {} with {} and {}\n'.format(self.url, self.username, self.password)
+#        return (
+#            'Would log in to {} with {} and {}\n'.format(self.url, self.username, self.password),
+#            'text/plain',
+#        )
 
 
 #class CaseFilter(filters.FilterBase):
--- a/lib/urlwatch/handler.py
+++ b/lib/urlwatch/handler.py
@@ -75,7 +75,12 @@ class JobState(object):
         try:
             try:
                 self.load()
-                data, content_type = self.job.retrieve(self)
+                retrieved = self.job.retrieve(self)
+                try:
+                    data, content_type = retrieved
+                except ValueError:  # backward compatibility
+                    data = retrieved
+                    content_type = ''
                 if content_type:
                     self.job.content_type = content_type
                     self.job.mime_type = content_type.split(';')[0].strip()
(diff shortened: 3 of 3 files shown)


lib/urlwatch/jobs.py
```diff
@@ -254,7 +254,7 @@ def retrieve(self, job_state):
         file_scheme = 'file://'
         if self.url.startswith(file_scheme):
             logger.info('Using local filesystem (%s URI scheme)', file_scheme)
-            return open(self.url[len(file_scheme):], 'rt').read()
+            return (open(self.url[len(file_scheme):], 'rt').read(), 'text/plain')
```
thp: The local file doesn't necessarily need to be `text/plain`, could be a different type?

lib/urlwatch/jobs.py
```diff
@@ -209,7 +209,7 @@ class UrlJob(Job):
     __required__ = ('url',)
     __optional__ = ('cookies', 'data', 'method', 'ssl_no_verify', 'ignore_cached', 'http_proxy', 'https_proxy',
                     'headers', 'ignore_connection_errors', 'ignore_http_error_codes', 'encoding', 'timeout',
-                    'ignore_timeout_errors', 'ignore_too_many_redirects')
+                    'ignore_timeout_errors', 'ignore_too_many_redirects', 'content_type', 'mime_type')
```
thp: What's the difference between `content_type` and `mime_type` here?

lib/urlwatch/jobs.py
```diff
@@ -358,4 +364,4 @@ def retrieve(self, job_state):
         from requests_html import HTMLSession
         session = HTMLSession()
         response = session.get(self.navigate)
-        return response.html.html
+        return (response.html.html, response.headers.get('Content-type', ''))
```
thp: Should there be a default fallback value for `response.headers.get()`? Maybe `text/html`?

share/urlwatch/examples/hooks.py.example
```diff
@@ -41,7 +41,10 @@ from urlwatch import reporters
 #    __required__ = ('username', 'password')
 #
 #    def retrieve(self, job_state):
-#        return 'Would log in to {} with {} and {}\n'.format(self.url, self.username, self.password)
+#        return (
+#            'Would log in to {} with {} and {}\n'.format(self.url, self.username, self.password),
```
thp: Unrelated change, please remove.

lib/urlwatch/jobs.py
tomaszn: Could be configured manually per job. And if not defined per job, then fallback could be taken from settings. If no fallback in settings, then libmagic can be used.

lib/urlwatch/jobs.py
tomaszn: - `content_type` is the raw response header including optional encoding information.
 - `mime_type` is the first part before optional ";" and doesn't contain optional encoding information, which is already taken into account when decoding the response. this is what you'd usually use.

lib/urlwatch/jobs.py
tomaszn: A typical web server would have that header. If it is missing, then the server is probably a custom script. Then maybe `text/plain`? or using [magic.from_buffer()](https://github.com/messense/magicfile#usage) to determine it?

share/urlwatch/examples/hooks.py.example
tomaszn: This pull request changes `JobState.process()` so that it requires all `retrieve` implementations to return a tuple:
```python
  data, content_type = self.job.retrieve(self)
```
So this example needs updating unless backward compatibility is implemented in `JobState.process()`.

lib/urlwatch/jobs.py
thp: These are settings from the jobs list YAML, you don't need to specify them here? (users are not supposed to "set" those values in the job configuration, right?)

lib/urlwatch/jobs.py
thp: Yes, make it an `__optional__` setting in the Job (it make sense for file jobs), maybe `mimetype` or something and if not set, use `text/plain`.

lib/urlwatch/jobs.py
thp: Yes, `text/plain` is probably also fine. However, how does that deal with headers like:

    Content-Type: text/html; charset=utf-8

share/urlwatch/examples/hooks.py.example
thp: Oh yeah true.. That's kind of unfortunate, it would by default break all custom user scripts. I would make it so that `self.job.retrieve(self)` can handle both the return value being a string (legacy behavior) or the return value being a 2-tuple. But let's think a bit more about this (I'll add another PR review comment).

lib/urlwatch/jobs.py
thp: Ok, nevermind, I understood how you plan to use this now... Basically that's a kind of "separate" kind of key that we (I think) haven't had yet where we want the job to obtain those properties (so it can be used for the filter matching) but it's not something we want the user to specify via the config variable (we'd otherwise overwrite it?).

lib/urlwatch/jobs.py
tomaszn: It's returned in a tuple and stored as `content_type`. Additionally, `text/html` is extracted and stored as `mime_type`.

lib/urlwatch/jobs.py
tomaszn: Right, I included them here so the automatic matching takes these two dictionary keys in `MATCH` into account.

After your remarks about defaults in `ShellJob`s maybe it should be possible to configure it (one can create say screenshots with a script).

tomaszn:
Thanks for the review.

1. Including all headers. Might be useful. But some headers are numeric, like `Content-Length`, and to make use of them the matching algorithm should support comparison operators like "greater then" etc. Are you thinking about returning a dict from `retrieve`?

2. The `charset` parameter has been somehow handled (look for `CHARSET_RE` in jobs.py), but its value is ignored. The decoding is done by trying several combinations of parameters, ignoring the charset value. I wrapped that in a function definition, so it can be later worked on and tested:
    ```python
        def try_decoding(response):
            try:
                try:
                    try:
                        return response.content.decode('utf-8')
                    except UnicodeDecodeError:
                        return response.content.decode('latin1')
                except UnicodeDecodeError:
                    return response.content.decode('utf-8', 'ignore')
            except LookupError:
                # If this is an invalid encoding, decode as ascii (Debian bug 731931)
                return response.content.decode('ascii', 'ignore')
    ```
    I think that it even might work as intended, because `charset` can be configured incorrectly. Anyway, `mime_type` is extracted from the header to not care about charset.

3. Please also take a look at #434. It introduces `needs_binary`, which is a "binary" version of `mime_type`. I think that implementing a PDF conversion filter with my approach would be simpler.

4. I think that the matching mechanism is too simple. Let's take a look at a case of user, that just wants to have hooks for HTML and Atom:
   - AtomHook:    `MATCH = {'mime_type': 'application/atom+xml'}`
   - AtomHook2:  `MATCH = {'url': re.compile('.*/atom.xml')}`, because some his favorite blogs have misconfigured `Content-Type` and use `text/html`
   - HtmlHook: `MATCH = {'mime_type': 'text/html'}`

    Currently both matching hooks will be used for these misconfigured `atom.xml` urls, which is not what the user wants. I see two solutions for this, which are not exclusive:
    - Introducing logical operators (AND, OR, NOT). A quick search reveals a library that implements necessary expressions: https://pypi.org/project/filtration/
    -  Priorities. If a hook matches, then lower priority hooks (fallbacks) are not processed. It should also allow setting the order in which hooks in one priority are run. So a hook for say `cnn.com` would run first, and then a default HTML processing would take place (or other way round, but in a predictable manner).

lib/urlwatch/jobs.py
tomaszn: Oh, I forgot that empty `content_type` is also acceptable, and in that case `job.content_type` and `job.mime_type` are not created. Maybe that's the way to go, rather than guess?

thp:
> * Including all headers. Might be useful. But some headers are numeric, like `Content-Length`, and to make use of them the matching algorithm should support comparison operators like "greater then" etc. Are you thinking about returning a dict from `retrieve`?

For the "comparison operators", you could just have a lambda:
```python
class AtomFeedFilter(AutoMatchFilter):
    MATCH = {'Content-Length': lambda value: int(value) > 300}
```

thp:
..and there's always the possibility of just writing a custom `match()` method for filters where it's complicated (e.g. Atom feed based on content-type, but then also based on URL, because of misconfigured server).

If just looking at the headers, this is probably fine, and things like "Content-Type: text/html; charset=utf-8" can be matched using the lambda matcher (`lambda value: 'text/html' in value`)

thp:
Not 100% sure if this is a good idea or not, but let's see what you come up with (and the "generic" HTTP header matching I'd feel more comfortable than this use case of trying to match Atom feeds).

thp:
@tomaszn Any updates on this?

thp:
Closing for now. Please rebase/reopen if you want to see this happening still.

PR #461, opened 2020-03-04
Add beautifier filter

agix:
To improve grep or grepi usage (at least) I found it usefull to beautify html/js/css so `diff` can work better on minified website.

I added a Dockerfile too and it should be integrated in this project. I can do another PR if you prefer.


obbardc:
+1 on the Dockerfile

agix:
Sure sorry I was out last week. I'm splitting it now.

lib/urlwatch/filters.py
```diff
@@ -155,6 +159,26 @@ def filter(self, data, subfilter=None):
             return data
 
 
+class BeautifierFilter(FilterBase):
+    """Beautify HTML"""
+
+    __kind__ = 'beautifier'
```
thp: Let's call this `beautify`, to be in line with the "verb" style of other filters (e.g. `strip` and not `stripper`, `format-json` and not `json-formatter`). Also change the class name to `BeautifyFilter` (or even `beautify-html` + `BeautifyHTMLFilter`?).

README.md
```diff
@@ -35,7 +35,7 @@ urlwatch 2 requires:
 
 The dependencies can be installed with (add `--user` to install to `$HOME`):
 
-`python3 -m pip install pyyaml minidb requests keyring appdirs lxml cssselect`
+`python3 -m pip install pyyaml minidb requests keyring appdirs lxml cssselect beautifulsoup4 jsbeautifier cssbeautifier`
```
thp: Can/should we make those optional, so that we don't pull in all those modules by default?
Import them only inside the `filter()` method of the beautify class.

See also "Optional dependencies" below.

lib/urlwatch/filters.py
agix: As it beautify html, js and css, I'll stick to `Beautify`.

lib/urlwatch/filters.py
```diff
@@ -155,6 +156,30 @@ def filter(self, data, subfilter=None):
             return data
 
 
+class BeautifyFilter(FilterBase):
+    import jsbeautifier
```
thp: It does not make sense to do this in the class context, you have to do it inside the function code (inside the block that starts with `def filter(...):`

agix:
oh shit sorry... I wasn't concentrated... I'll change this asap :)

Le mer. 25 mars 2020 à 21:09, Thomas Perl <notifications@github.com> a
écrit :

thp:
Thanks!

PR #472, opened 2020-04-06
Create requirements file for easier installation

naclcaleb:
I added a `requirements.txt` file so the dependencies can be installed simply with: `pip install -r requirements.txt`


thp:
Thanks!

PR #476, opened 2020-04-23
Adding sort filter

bretterism:
This is to add a sorting option in the case that a web page has its results in different order with each refresh.


thp:
Cool, thanks!

PR #479, opened 2020-04-25
Add a new filter reverse inspired by sort.

uyfv554:
reverse the order of iterms.

### DIFF
--- a/README.md
+++ b/README.md
@@ -223,6 +223,14 @@ url: https://example.net/
 filter: sort
 ```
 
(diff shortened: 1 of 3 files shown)


mborsetti:
Great idea, but wouldn't it be better if it were an option of the existing sort filter, i.e. `filter/sort/reverse: true`

uyfv554:
What If one attempts to reverse a list rather than sort with a inverse order?

PR #481, opened 2020-05-05
Remove pycodestyle hard dep

polyzen:
pycodestyle is only used during testing, but after the switch to entry
points, urlwatch will not run without it.

For reference: https://bugs.archlinux.org/task/66532

Since you've added a requirements.txt and Travis uses that by default,
may as well use that in .travis.yml.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,5 @@
 language: python
+cache: pip
 python:
   - "3.5"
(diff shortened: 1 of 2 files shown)


polyzen:
Should the `distutils` import and `InstallDependencies` class be removed as well?

PR #482, opened 2020-05-05
Docs: rewrote Jobs page and added an Introduction page

mborsetti:
Did a fair amount of cleanup to a small part of the documentation (the Jobs page) and started an Introduction page.  

Of note is the deprecation of the `kind` key, which was extremely confusing to this new user and has been made redundant (per [current documentation](https://urlwatch.readthedocs.io/en/latest/jobs.html): "`kind` [...] is usually auto-detected and can be left out, because the job kind can be derived from the unique key)" and therefore unneeded.

Also moved from Jobs to Advanced Topics section the tip about "If you are watching pages that change seldomly, but you still want to be notified daily if urlwatch still works, you can watch the output of the date command" 

Finally, added reST labels to allow cross-referencing where needed (more need to be added) and made opportunistic fixes such as using standard reStructuredText directive for yaml [code-blocks](https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-code-block).

### DIFF
--- a/docs/source/advanced.rst
+++ b/docs/source/advanced.rst
@@ -1,10 +1,12 @@
+.. _advanced_topics:
+
 Advanced Topics
 ===============
 
 If you want to change some settings for all your jobs, edit the
 ``job_defaults`` section in your config file:
 
-.. code:: yaml
+.. code-block:: yaml
 
    ...
    job_defaults:
@@ -25,7 +27,6 @@ Quickly adding new URLs to the job list from the command line::
     urlwatch --add url=http://example.org,name=Example
 
 
-
 Using word-based differences
 ----------------------------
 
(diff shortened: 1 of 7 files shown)


cfbao:
I have mixed feeling about the removal of `kind` in docs. For basic usage, it probably makes sense to simplify the introduction. But the docs now looks rather misleading/unhelpful for advanced usage, especially for people intent on implementing custom jobs in `hooks.py`, which can easily break the auto-detection of `kind`.

I think an advanced section on the true underlying mechanism of `kind` is in dire need.

mborsetti:
> I have mixed feeling about the removal of `kind` in docs. For basic usage, it probably makes sense to simplify the introduction. But the docs now looks rather misleading/unhelpful for advanced usage, especially for people intent on implementing custom jobs in `hooks.py`, which can easily break the auto-detection of `kind`.
> 
> I think an advanced section on the true underlying mechanism of `kind` is in dire need.

Barely knew that `hooks.py` exists, as there's no reference to it in the docs, and have absolutely zero  clue on what it does or how to use it. 

Can you contribute some documentation?

PR #484, opened 2020-05-07
Have html2text filter using Python library return full ("absolute") URLs instead of relative ones

mborsetti:
The `html2text` filter `method: pyhytml2text` (which uses the `html2text` python library ) will now return full ("absolute") URLs instead of relative ones.

The implementation is scalable:

1. `handler.JobState.process` extracts the URL as `baseurl` and passes it to `filters.FilterBase.process`
2. `filters.FilterBase.process` passes it on to the `filtercls` classes, so all filters have access to it
3. The filter `html2text` passes it to `html2txt.html2text`
4. `html2txt.html2text`, when calling the `html2text` Python library, now sets the library's `baseurl` option so it can correctly parse relative URLs

Benefit: links that previously read e.g. `../../index.html` now read  e.g. `https://www.example.com/index.html`

### DIFF
--- a/lib/urlwatch/filters.py
+++ b/lib/urlwatch/filters.py
@@ -200,7 +200,8 @@ class Html2TextFilter(FilterBase):
             method = subfilter
             options = {}
         from .html2txt import html2text
-        return html2text(data, method=method, options=options)
+        return html2text(data, baseurl=getattr(self.job, 'url', getattr(self.job, 'navigate', '')),
+                         method=method, options=options)
 
 
 class Pdf2TextFilter(FilterBase):
--- a/lib/urlwatch/html2txt.py
+++ b/lib/urlwatch/html2txt.py
@@ -48,7 +48,7 @@ def html2text(data, baseurl, method, options):
                         options: https://linux.die.net/man/1/html2text
      'bs4'            - Use Beautiful Soup library to prettify the HTML
                         options: "parser" only, bs4 supports "lxml", "html5lib", and "html.parser"
-                        http://beautiful-soup-4.readthedocs.io/en/latest/#specifying-the-parser-to-use
+                        https://www.crummy.com/software/BeautifulSoup/bs4/doc/#specifying-the-parser-to-use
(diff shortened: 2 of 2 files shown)


lib/urlwatch/html2txt.py
```diff
@@ -61,14 +61,15 @@ def html2text(data, method, options):
...
+        parser.baseurl = baseurl
         for k, v in options.items():
             setattr(parser, k.lower(), v)
         d = parser.handle(data)
         return d
 
     if method == 'bs4':
         from bs4 import BeautifulSoup
-        parser = options.pop('parser', 'html.parser')
+        parser = options.pop('parser', 'lxml')
```
cfbao, line 72: unrelated change?

lib/urlwatch/filters.py
```diff
@@ -82,17 +82,17 @@ def auto_process(cls, state, data):
...
         if filtercls is None:
             raise ValueError('Unknown filter kind: %s:%s' % (filter_kind, subfilter))
-        return filtercls(state.job, state).filter(data, subfilter)
+        return filtercls(state.job, state).filter(data, baseurl, subfilter)
 
     def match(self):
         return False
 
-    def filter(self, data, subfilter=None):
+    def filter(self, data, baseurl, subfilter=None):
```
cfbao, line 95: `baseurl` sounds too specific a name for the base class. While other filters might also utilize the job url, in their context it probably doesn't make sense to call it "base" url.

To make it even more generic, I wonder if it makes sense to pass the entire `job` object into the filter, and let filters decide for themselves what they want to do. I can't recall exactly what, but I had situations where it'd be useful to have `job` in filters.

mborsetti:
> in fact, `FilterBase` sets `self.job`, so you can inspect the job object from within each filter.

Was looking for something like that but somehow completely missed it.

I'll try to rework it.  Thanks!

mborsetti:
So much easier!  And I even think I have sort-of-figured out git. :)

Thanks so much.

thp:
Thanks!

PR #490, opened 2020-05-09
Enabled sub-keys for browser/navigate jobs

mborsetti:
I tried to pass a proxy (and later headers) to a `navigate` job (which uses a requests_html headless browser, very useful!) but ... there was no code to make it happen!

Turns out that `requests_html headless.HTMLSession()` inherits` requests.Session()`, so it was quite easy to code that in: it's mostly the same as job kind `url` with some minor changes.

And yes, it works! :)

### DIFF
--- a/lib/urlwatch/jobs.py
+++ b/lib/urlwatch/jobs.py
@@ -321,7 +321,7 @@ class UrlJob(Job):
 
     def add_custom_headers(self, headers):
         """
-        Adds custom request headers from the job list (URLs) to the pre-filled dictionary `headers`.
+        Adds custom request headers from the job list to the pre-filled dictionary `headers`.
         Pre-filled values of conflicting header keys (case-insensitive) are overwritten by custom value.
         """
         headers_to_remove = [x for x in headers if x.lower() in [y.lower() for y in self.headers]]
@@ -361,6 +361,9 @@ class BrowserJob(Job):
     __kind__ = 'browser'
 
     __required__ = ('navigate',)
+    __optional__ = ('cookies', 'data', 'ssl_no_verify', 'ignore_cached', 'http_proxy', 'https_proxy',
+                    'headers', 'ignore_connection_errors', 'ignore_http_error_codes', 'encoding', 'timeout',
+                    'ignore_timeout_errors', 'ignore_too_many_redirects')
 
     LOCATION_IS_URL = True
 
@@ -368,7 +371,68 @@ class BrowserJob(Job):
         return self.navigate
 
     def retrieve(self, job_state):
+        headers = {
+            'User-agent': urlwatch.__user_agent__,
+        }
+
+        proxies = {
+            'http': os.getenv('HTTP_PROXY'),
+            'https': os.getenv('HTTPS_PROXY'),
+        }
+
+        if job_state.etag is not None:
+            headers['If-None-Match'] = job_state.etag
+
+        if job_state.timestamp is not None:
+            headers['If-Modified-Since'] = email.utils.formatdate(job_state.timestamp)
+
+        if self.ignore_cached or job_state.tries > 0:
+            headers['If-None-Match'] = None
+            headers['If-Modified-Since'] = email.utils.formatdate(0)
+            headers['Cache-Control'] = 'max-age=172800'
+            headers['Expires'] = email.utils.formatdate()
+
+        if self.http_proxy is not None:
+            proxies['http'] = self.http_proxy
+        if self.https_proxy is not None:
+            proxies['https'] = self.https_proxy
+
+        file_scheme = 'file://'
+        if self.navigate.startswith(file_scheme):
+            logger.info('Using local filesystem (%s URI scheme)', file_scheme)
+            from requests_html import HTML
+            return HTML(open(self.navigate[len(file_scheme):], 'rt').read())
+
+        if self.headers:
+            self.add_custom_headers(headers)
+
+        if self.timeout is None:
+            # default timeout
+            timeout = 60
+        elif self.timeout == 0:
+            # never timeout
+            timeout = None
+        else:
+            timeout = self.timeout
+
         from requests_html import HTMLSession
         session = HTMLSession()
-        response = session.get(self.navigate)
+        response = session.get(url=self.navigate,
+                               data=self.data,
+                               headers=headers,
+                               verify=(not self.ssl_no_verify),
+                               cookies=self.cookies,
+                               proxies=proxies,
+                               timeout=timeout)
         return response.html.html
+
+    def add_custom_headers(self, headers):
+        """
+        Adds custom request headers from the job list to the pre-filled dictionary `headers`.
+        Pre-filled values of conflicting header keys (case-insensitive) are overwritten by custom value.
+        """
+        print(headers)
+        headers_to_remove = [x for x in headers if x.lower() in [y.lower() for y in self.headers]]
+        for header in headers_to_remove:
+            headers.pop(header, None)
+        headers.update(self.headers)


lib/urlwatch/jobs.py
```diff
@@ -361,14 +361,78 @@ class BrowserJob(Job):
...
+                               proxies=proxies,
+                               timeout=timeout)
         return response.html.html
+
+    def add_custom_headers(self, headers):
+        """
+        Adds custom request headers from the job list to the pre-filled dictionary `headers`.
+        Pre-filled values of conflicting header keys (case-insensitive) are overwritten by custom value.
+        """
+        print(headers)
```
thp, line 434: Remove this `print()`

lib/urlwatch/jobs.py
```diff
@@ -361,14 +361,78 @@ class BrowserJob(Job):
...
+        headers_to_remove = [x for x in headers if x.lower() in [y.lower() for y in self.headers]]
+        for header in headers_to_remove:
+            headers.pop(header, None)
```
thp, line 437: Can be `del headers[header]`

lib/urlwatch/jobs.py
```diff
@@ -361,14 +361,78 @@ class BrowserJob(Job):
...
+        response = session.get(url=self.navigate,
+                               data=self.data,
+                               headers=headers,
+                               verify=(not self.ssl_no_verify),
+                               cookies=self.cookies,
+                               proxies=proxies,
+                               timeout=timeout)
         return response.html.html
+
+    def add_custom_headers(self, headers):
```
thp, line 429: Should the adding of custom headers be a separate PR? Can you add some unit tests for verifying that it works and does the right thing (replace keys with a different case, etc..)?

Also, there is a `CaseInsensitiveDict` in `requests.structures` that we could use, since we anyway depend on `requests`.

lib/urlwatch/jobs.py
```diff
@@ -361,14 +361,78 @@ class BrowserJob(Job):
     __kind__ = 'browser'
 
     __required__ = ('navigate',)
+    __optional__ = ('cookies', 'data', 'ssl_no_verify', 'ignore_cached', 'http_proxy', 'https_proxy',
+                    'headers', 'ignore_connection_errors', 'ignore_http_error_codes', 'encoding', 'timeout',
+                    'ignore_timeout_errors', 'ignore_too_many_redirects')
```
thp, line 366: Can these optional keys be shared between the URL job and the Browser job? If nothing else, move the tuple of optionals to a named global and use it in both places, but maybe even a common ancestor class is doable?

lib/urlwatch/jobs.py
mborsetti: Ooops!

lib/urlwatch/jobs.py
mborsetti: No changes were made to the code copied from the URL job:
https://github.com/thp/urlwatch/blob/4654eb984dae6e1d7734fefcb3ae976f2d3d0843/lib/urlwatch/jobs.py#L327-L330

lib/urlwatch/jobs.py
mborsetti: No changes were made to the code copied from the URL job:
https://github.com/thp/urlwatch/blob/4654eb984dae6e1d7734fefcb3ae976f2d3d0843/lib/urlwatch/jobs.py#L321-L322

lib/urlwatch/jobs.py
mborsetti: URL also accepts the method key which Browser does not;

https://github.com/thp/urlwatch/blob/4654eb984dae6e1d7734fefcb3ae976f2d3d0843/lib/urlwatch/jobs.py#L217-L219

mborsetti:
Other than removing the errand `print()`, the other suggested changes pertain to modifications to existing code or to re architecting/refactoring (which I am not qualified to do).  Please advise!

lib/urlwatch/jobs.py
mborsetti: In rethinking this, my initial answer was incorrect as the Browser job (navigate key) can and should accept `method `as well, so the entirety of the code handling sub-keys is identical for both URL and Browser.

But am still quite confused as to why both `kind` **and** the key containing the URL changes between URL and Browser.

A clearer way (from an end user perspective) could be to eliminate the `kind` and simply look at the key: if `url` is found, then `requests `is used; if `navigate` is found, then `requests_html` is used.  Alternatively, the name of the key pointing to an URL doesn't change between the two types of jobs (`url` always points to the URL, and `kind` effectively becomes the _method_ of extracting data from that URL).

These are architecture and UX decisions.

cfbao: > A clearer way (from an end user perspective) could be to eliminate the kind and simply look at the key: if url is found, then requests is used; if navigate is found, then requests_html is used.

Personally I feel the exact opposite. It's much clearer to me to have a `kind` parameter to indicate the job type, and other parameters to specify the details. Combining different responsibilities (indicate type & specify details) into one parameter is much more confusing.

cfbao:
@mborsetti  before doing more work on this, you should be aware of existing issues with "browser" job:
* #293
* https://github.com/thp/urlwatch/pull/298#issuecomment-435691621

Currently, it doesn't actually launch a browser to do any JavaScript rendering. It is functionally identical to using `requests` directly.

It may seem like we can pass the same set of parameters to both "url" and "browser" jobs, but many (perhaps most) of these parameters (e.g. "method") would be functionally lost if we were to actually render the webpage with a browser.

As I elaborated in https://github.com/thp/urlwatch/pull/298#issuecomment-435691621, `requests-html` doesn't bring much to the table for "browser" job, compared to directly using `pyppeteer`. On the contrary, it puts in a middle layer that actually inhibits our control over the browser behaviour. That's why I've been trying to work out a solution to rewrite "browser" job directly using `pyppeteer`: #355.

mborsetti:
@cfbao thanks for your comments!

> Currently, it doesn't actually launch a browser to do any JavaScript rendering. It is functionally identical to using `requests` directly, just like "url" job.

Hmmm, interesting because it's doing this rendering in my environment (which Job did not do) and the `requests_html` [documentation](https://pypi.org/project/requests-html/) specifically says **Full JavaScript support!** (emphasis theirs).  The package doesn't seem to be maintained, so maybe you're seeing something I'm not.

> It may seem like we can pass the same set of parameters to both "url" and "browser" jobs, but many (perhaps most) of these parameters (e.g. "method") would be functionally lost if we were to actually render the webpage with a browser.

According to https://github.com/psf/requests-html/issues/218 `requests_html.HTMLSession()` inherits `requests.Session()`, and since the URL job uses `requests.Session()` the sub-keys are quite literally exactly the same.  Also "method" is one of either GET or POST, and since many webpages are fetched with a POST, this method should be available to a browser-rendered webpage engine.

> As I elaborated in [#298](https://github.com/thp/urlwatch/pull/298#issuecomment-435691621), `requests-html` doesn't bring much to the table for "browser" job, compared to directly using `pyppeteer`. On the contrary, it puts in a middle layer that actually inhibits our control over the browser behaviour. That's why I've been trying to work out a solution to rewrite "browser" job directly using `pyppeteer`: #355.

I missed that! I completely agree that directly using `pyppeteer` directly instead of dragging along the overhead of `requests-html` is a better, but I contributed my fix as it is one that works today (well, at least in my environment).  How's that effort coming?  Can I lend a hand?

cfbao:
> it's doing this rendering in my environment

I'm curious what your exact example is, because the way it's written now, there's no way a browser can be launched.
Currently all jobs in urlwatch run entirely in background thread pool, but you can only launch a control a browser from the main thread. See [relevant Python docs](https://docs.python.org/3/library/signal.html).

> the requests_html documentation specifically says Full JavaScript support! (emphasis theirs)

It's supported, but "browser" job isn't using that feature.

> According to psf/requests-html#218 requests_html.HTMLSession() inherits requests.Session(), and since the URL job uses requests.Session() the sub-keys are quite literally exactly the same. Also "method" is one of either GET or POST, and since many webpages are fetched with a POST, this method should be available to a browser-rendered webpage engine.

This is correct, because up to this point, `requests_html.HTMLSession` is simply using `requests`, doing the exact same thing, **with no browser launched**. Once you call `render()`, which actually launches a browser and run JavaScript, most of these parameters are thrown out of the window.

> How's that effort coming? Can I lend a hand?

It's mostly a design issue. It works, but I'm not happy with how it works. You can take a look at the PR I linked (and a few older ones linked there). If you have any better idea, feel free to leave comments.

mborsetti:
> > it's doing this rendering in my environment
> 
> I'm curious what your exact example is, because the way it's written now, there's no way a browser can be launched.
> Currently all jobs in urlwatch run entirely in background thread pool, but you can only launch a control a browser from the main thread. See [relevant Python docs](https://docs.python.org/3/library/signal.html).

You are obviously correct.  My rendering had nothing to do with lack of Javascript but an unrelated proxy addition.

> > the requests_html documentation specifically says Full JavaScript support! (emphasis theirs)
> 
> It's supported, but "browser" job isn't using that feature.

So the documentation is obviously **incorrect**.   What is browser for?  What should the documentation read?

polyzen:
> So the documentation is obviously **incorrect**. What is browser for? What should the documentation read?
> 
> After struggling with implementing _urlwatch_ I de-facto volunteered to bring the docs to a useful status but am struggling if the existing documentation is wrong... (BTW have already documented a big chunk of filters but am waiting to finish it before contributing it)

Seems like the docs should be correct once #355 is done, but may need some editing, at the least to `s/requests-html/pyppeteer/`.

Documentation revamping is always greatly appreciated, at least for myself from a user/packager standpoint. Also better docs means less issues I need to field or forward upstream. ;)

cfbao:
> What is browser for?

It was _supposed_ to do what the docs says and what you thought it did - launching a browser to do JavaScript rendering. But unfortunately it was just incorrectly implemented.

PR #491, opened 2020-05-10
Add a redis cache backend

scottmac:
I'm wanting to run multiple workers on different machines and also do some monitoring of changes. This allows the cache to live off box if needed.

The redis list type maps well to the cache as we only ever push new history snapshots and read from the beginning.

I saw @liewegas had written a postgres version in a fork so might be more than just me after this.

Also not sure how much further to take this, I could make the yaml bit a litlte more generic and throw together a MySQL one.

### DIFF
--- a/docs/source/advanced.rst
+++ b/docs/source/advanced.rst
@@ -146,3 +146,13 @@ of the ``date`` command, for example:
 
 Since the output of ``date`` changes every second, this job should produce a
 report every time urlwatch is run.
+
+
+Using Redis as a cache backend
+------------------------------------------
+If you want to use Redis as a cache backend over the default SQLite3 file::
+
+    urlwatch --cache=redis://localhost:6379/
+    
+There is no migration path from the SQLite3 format, the cache will be empty
+the first time Redis is used.0
--- a/docs/source/advanced.rst
+++ b/docs/source/advanced.rst
@@ -153,6 +153,6 @@ Using Redis as a cache backend
 If you want to use Redis as a cache backend over the default SQLite3 file::
 
     urlwatch --cache=redis://localhost:6379/
-    
+
 There is no migration path from the SQLite3 format, the cache will be empty
-the first time Redis is used.0
+the first time Redis is used.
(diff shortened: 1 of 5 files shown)


docs/source/advanced.rst
```diff
@@ -146,3 +146,13 @@ of the ``date`` command, for example:
...
+
+
+Using Redis as a cache backend
+------------------------------------------
+If you want to use Redis as a cache backend over the default SQLite3 file::
+
+    urlwatch --cache=redis://localhost:6379/
+    
+There is no migration path from the SQLite3 format, the cache will be empty
+the first time Redis is used.0
```
thp, line 158: Stray `0` at the end, please remove.

lib/urlwatch/cli.py
```diff
@@ -100,7 +100,12 @@ def main():
 
     # setup storage API
     config_storage = YamlConfigStorage(command_config.config)
-    cache_storage = CacheMiniDBStorage(command_config.cache)
+
+    if command_config.cache.startswith('redis://') or command_config.cache.startswith('rediss://'):
```
thp, line 104: Could be:

    if any(command_config.cache.startswith(prefix) for prefix in ('redis://', 'rediss://')):

thp:
Merged, thanks!

lib/urlwatch/storage.py
```diff
@@ -507,3 +517,72 @@ def clean(self, guid):
...
+    def __init__(self, filename):
+        super().__init__(filename)
+
+        if redis is None or msgpack is None:
+            raise ImportError('redis + msgpack are missing')
+
+        self.db = redis.from_url(filename)
+
+    def _make_key(self, guid):
+        return 'guid:' + guid
```
cfbao, line 532: @scottmac why do you prepend `'guid: '` to the key? Is it purely to make it look nicer, or is there a technical reason?

I'm curious, as I'll probably need to make changes to all cache backends, and I'm not very familiar with redis.

Issue #148, opened 2017-04-28
Multiple URL lists


strauhmanis:
Hi Thomas!

I found myself in need of having separate URL lists so as to check different links at different time intervals.

The way that I can think of achieving this goal now, is to create a shell script or several of them and with their help to rename `urls.yaml` back and forth between `my-urls-list1.yaml`, `my-urls-list2.yaml`, etc.

Those who can, have also an option of using urlwatch as a library to create an elegant solution, as far as I understand this, but it would be nice if at some future point it were to be possible to supply urlwatch specific lists without having to come up with a gettting-around-it-script or an app.

Thanks!

thp:
Well the solution would probably be to have a new `interval` key in the YAML file and let urlwatch handle the different intervals with a single list?

With that said, there's also the `--urls` command line option that should do the trick?

    urlwatch --urls=my-urls-list1.yaml

thp:
To clarify, the `interval` key isn't yet implemented, but might be a good idea, but the `--urls` option is already implemented and should work already.

strauhmanis:
Awesome! Apparently I browsed urlwatch code not that thoroughly and missed that already-in-place feature.

For anyone else who would be reading this, remember to supply full path to the URLs list:

`urlwatch --urls=/Users/me/.urlwatch/my-urls-list1.yaml`

Many thanks Thomas!

strauhmanis:
I've been banging my head for a past hour trying to find an answer on why is a path to a URL list in a bash script is being "adjusted" and becomes invalid, while the same command in a terminal works fine:

This is the terminal output. First I run a script (see the script below), then I simply run a command:
```________________________________________________________________________________
| ~/Dropbox/Scripts/urlwatch @ TitanPonies (ervsts) 
| => bash urlwatch-go-Apple.sh

 in order to use urlwatch.ers/ervsts/Dropbox/Scripts/urlwatch/urls-apps-Apple.yaml
    Use "urlwatch --edit" to open the file with your editor.
    
________________________________________________________________________________
| ~/Dropbox/Scripts/urlwatch @ TitanPonies (ervsts) 
| => urlwatch --urls=/Users/ervsts/Dropbox/Scripts/urlwatch/urls-apps-Apple.yaml
________________________________________________________________________________
| ~/Dropbox/Scripts/urlwatch @ TitanPonies (ervsts) 
| => 
```

The script (if I may call it that) itself:
```
#!/bin/bash
[...]

PR #494, opened 2020-05-10
Reorganize migration code

cfbao:
* Move storage objects creation from `cli.main` to `main.Urlwatch.__init__`
* Move urls file check from `migration.migrate_urls` to `main.Urlwatch.__init__`
* Move migration function calls from `main.Urlwatch.__init__` to `cli.main`

Each bullet point corresponds to one self-contained commit.

Mostly refactoring in preparation for introducing new cache DB.
I feel this way the responsibilities of migration/initialization/validation are better separated/organized.

### DIFF
--- a/lib/urlwatch/cli.py
+++ b/lib/urlwatch/cli.py
@@ -53,6 +53,7 @@ if bindir != 'bin':
 from urlwatch.command import UrlwatchCommand
 from urlwatch.config import CommandConfig
 from urlwatch.main import Urlwatch
+from urlwatch.migration import migrate_cache, migrate_urls
 
 # Ignore SIGPIPE for stdout (see https://github.com/thp/urlwatch/issues/77)
 try:
@@ -92,6 +93,10 @@ def main():
     command_config = CommandConfig(pkgname, urlwatch_dir, bindir, prefix,
                                    config_file, urls_file, hooks_file, cache_file, False)
     setup_logger(command_config.verbose)
+
+    # migration
+    migrate_urls(command_config)
+    migrate_cache(command_config)
 
     # setup urlwatcher
     urlwatch = Urlwatch(command_config)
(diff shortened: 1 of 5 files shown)


lib/urlwatch/migration.py
```diff
@@ -81,8 +65,8 @@ def migrate_cache(urlwatcher):
         print("""
     Migrating cache: {cache_dir} -> {cache_db}
     """.format(cache_dir=cache_dir, cache_db=cache))
-
         old_cache_storage = CacheDirStorage(cache_dir)
-        urlwatcher.cache_storage.restore(old_cache_storage.backup())
-        urlwatcher.cache_storage.gc([job.get_guid() for job in urlwatcher.jobs])
+        new_cache_storage = CacheMiniDBStorage(cache)
+        new_cache_storage.restore(old_cache_storage.backup())
+        new_cache_storage.close()
```
cfbao, line 71: The only functional change in this PR is here. I removed the `gc` call during migration.
This way, users using multiple url files (example use case: #148) can also safely migrate their cache.

lib/urlwatch/cli.py
```diff
@@ -98,13 +98,12 @@ def main():
                                    config_file, urls_file, hooks_file, cache_file, False)
     setup_logger(command_config.verbose)
 
-    # setup storage API
-    config_storage = YamlConfigStorage(command_config.config)
-    cache_storage = CacheMiniDBStorage(command_config.cache)
-    urls_storage = UrlsYaml(command_config.urls)
+    # migration
+    migrate_urls(command_config)
+    migrate_cache(command_config)
```
thp, line 103: Previously before the migration, `check_directories()` was called, which created directories on demand -- this is now done AFTER migration, isn't that a problem?

cfbao: Looking at the logic of these two migration functions, I don't think this would be a problem. Migration happens only if `urls_txt`/`cache_dir` exist, but they are both under `urlwatch_dir` by definition.

In any case, I've added a commit that creates parent directory of `urls` and `cache` before the migration code creates these files.

cfbao:
Rebased to fix conflicts.

thp:

PR #495, opened 2020-05-11
Update hooks importing code

cfbao:
[`imp`](https://docs.python.org/3/library/imp.html) is deprecated.
New implementation is largely the same as the [example](https://docs.python.org/3/library/importlib.html#importing-a-source-file-directly) in `importlib`'s docs.

### DIFF
--- a/lib/urlwatch/util.py
+++ b/lib/urlwatch/util.py
@@ -33,6 +33,9 @@ import os
 import platform
 import subprocess
 import shlex
+import importlib.machinery
+import importlib.util
+import sys
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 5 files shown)


lib/urlwatch/util.py
```diff
@@ -99,3 +102,12 @@ def edit_file(filename):
         raise SystemExit('Please set $VISUAL or $EDITOR.')
 
     subprocess.check_call(shlex.split(editor) + [filename])
+
+
+def import_module_from_source(module_name, source_path):
+    loader = importlib.machinery.SourceFileLoader(module_name, source_path)
+    spec = importlib.util.spec_from_file_location(module_name, source_path, loader=loader)
```
cfbao, line 109: Explicitly specifying a loader to allow arbitrary file extension.
E.g. `share/urlwatch/examples/hooks.py.example`
Otherwise `spec` would be `None` on unsupported extensions.

thp:
Thanks!

PR #498, opened 2020-05-16
'splitlines([keepends])'

mborsetti:
Replaced convoluted and slow list comprehension with splitlines(keepends=True) within TextReporter

Made language more pythonic in similar splitlines() statement within HtmlReporter


mborsetti:
Not sure why it's failing on a file I did not make any changes to so am not touching it.

PR #500, opened 2020-05-17
Add mechanism to request/release job resources

cfbao:
Another new design in preparation for re-implementation of browser job. I feel much happier about this one.

Similar to #355, added `request_resources` and `release_resources` to `JobBase`.
But this time, `JobState` becomes a context manager (inspired by https://github.com/thp/urlwatch/pull/355#discussion_r422605448). It wraps the request/release method calls so exceptions can be captured/reported without crashing the entire run.
The `resources` dict is now better encapsulated as a class variable of `JobState`.

I'm not turning `JobBase` into a context manager, because I still don't want job objects to carry state. I quite like the separation of definition & state between job and job state classes.

To reduce PR size and ease review, this PR doesn't (or at least shouldn't) introduce any functional changes. Real browser job implementation will be in another PR, if this one's approved.

May be desirable to enable "hide whitespace changes" because there are indentation changes.

### DIFF
--- a/lib/urlwatch/handler.py
+++ b/lib/urlwatch/handler.py
@@ -70,10 +70,8 @@ class JobState(object):
         try:
             self.job.release_resources(JobState.resources)
         except Exception as ex:
-            logger.error('Got exception while releasing resources for job: %r', self.job, exc_info=ex)
-            if self.exception is None:
-                self.exception = ex
-                self.traceback = traceback.format_exc()
+            # Log warning only. We don't want exceptions from releasing resources to override job run results
+            logger.warning('Got exception while releasing resources for job: %r', self.job, exc_info=ex)
 
     def load(self):
         guid = self.job.get_guid()
--- a/lib/urlwatch/handler.py
+++ b/lib/urlwatch/handler.py
@@ -41,6 +41,8 @@ logger = logging.getLogger(__name__)
 
 
 class JobState(object):
+    resouces = {}
+
     def __init__(self, cache_storage, job):
         self.cache_storage = cache_storage
         self.job = job
@@ -54,6 +56,24 @@ class JobState(object):
         self.tries = 0
         self.etag = None
         self.error_ignored = False
+
+    def __enter__(self):
+        try:
+            self.job.request_resources(JobState.resouces)
+        except Exception as ex:
+            self.exception = ex
+            self.traceback = traceback.format_exc()
+        finally:
+            return self
+
+    def __exit__(self, exc_type, exc_value, traceback):
+        try:
+            self.job.release_resources(JobState.resouces)
+        except Exception as ex:
+            logger.error('Got exception while releasing resources for job: %r', self.job, exc_info=ex)
+            if self.exception is None:
+                self.exception = ex
+                self.traceback = traceback.format_exc()
 
     def load(self):
         guid = self.job.get_guid()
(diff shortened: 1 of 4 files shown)


lib/urlwatch/handler.py
```diff
@@ -55,6 +57,24 @@ def __init__(self, cache_storage, job):
...
+            return self
+
+    def __exit__(self, exc_type, exc_value, traceback):
+        try:
+            self.job.release_resources(JobState.resouces)
+        except Exception as ex:
+            logger.error('Got exception while releasing resources for job: %r', self.job, exc_info=ex)
+            if self.exception is None:
+                self.exception = ex
+                self.traceback = traceback.format_exc()
```
cfbao, line 76: Error handling here is a little tricky. The job run may be successful during `process()` call, and produced useful results that users would definitely want to see in reports, but we may still want users to be aware of errors happened here. How should we report this?

lib/urlwatch/handler.py
```diff
@@ -41,6 +41,8 @@
 
 
 class JobState(object):
+    resouces = {}
```
thp, line 44: Typo: `resources`

thp: Also, are we sure it should be a class member (one per class, not one per instance)?

lib/urlwatch/handler.py
thp: Maybe just a `logger.warn()` or something?

cfbao: I suspect many users are not seeing most logs, because urlwatch would run in the background by a scheduler. But still, this would be preferable to hiding diff results.

lib/urlwatch/handler.py
cfbao: In the design I have in mind, yes, it should be a class member, i.e. it's a globally unique static variable.
The point of it is to share expensive resources among different jobs. In browser job's case, these are the browser instance and async event loop.

cfbao:
> If this is just for `browser` jobs, maybe just have a global object that takes care of managing the shared resource, then this wouldn't be needed at all?

Some of my considerations with the current design, mostly centred around customizability/extensibility:

* As the browser/navigate job type gets more developed in the future, I can see urlwatch allowing users to specify _how_ the browser is launched, e.g. which Chrome/Chromium binary and which profile to use. This can ease the monitoring of web pages behind logins, because users can simply let urlwatch use the Chrome profile they normally use, without manually fiddling with cookies.
If this is to be the case, I think it makes sense to keep the browser launching process close to job/jobstate for easy access to job configs. Users can even choose to launch a few differently configured browser processes (but still fewer than one per job) with this design.

* Hooks
For users to use hooks to freely subclass/modify or even implement their own version of browser jobs, all functionalities need be exposed via the job class, without any "magic" that's only accessible in urlwatch's core framework.

* Minimum disruption to "framework" code
This has been on my mind since your early [comment](https://github.com/thp/urlwatch/pull/298#issuecomment-436025040) (almost two years ago! wow) about keeping `worker.py` changes to a minimum. To achieve that, I feel it's best to not give any job type (or filter type) any special treatment, but keep everything generic, which is also good for clean & readable framework code and future extensiblity.

thp:
Inspired by your PR, this is what I came up with in #546:

https://github.com/thp/urlwatch/pull/546/commits/004993ae7455146be67dbfd36714d80c10bb5b8b

It's basically the exact same code/logic, except that:

- There's no `JobState.resources` dict that's passed around (job needs to manage it elsewhere)
- Renamed functions on `JobBase` to `main_thread_enter()` and `main_thread_exit()`

The only downside might be extensibility, but we could always make the `urlwatch.browser` (see #546) module more open for extensibility.

PR #512, opened 2020-06-24
Add ifttt reporter

agix:
Add reporter ifttt (https://ifttt.com/) with `webhook_url` configuration.


thp:
This needs documentation on how to set up / how to use / how to test (in `docs/source/reporters.rst`).

thp:
Thanks, I've merged it now but decided to improve the documentation a bit and instead of requiring the full webhook URL, only require the event name and the key to be configured, als some other details (e.g. `value1` now only contains the verb) I hope this is okay for you.

agix:
Perfect thank you !

PR #518, opened 2020-07-06
Significant upgrades to HTML reporting

mborsetti:
Significant upgrades to HTML reporting:
* Markdown is reconstructed into HTML 
* HTML output now survives stylesheet rewriting by email clients

Major changes:
* Filters that output Markdown can opt to have the Markdown reconstructed to HTML if unified diffing (default) and the `markdown2` library is found. This preserves inline links (turning them into clickable ones) and all other basic formatting such as lists, bolding, italics, etc. This upgrade makes HTML reporting radically more legible and useful.
  - `job.is_markdown` key is set to True by a filter to indicate it outputs Markdown.
* The reconstruction of Markdown to HTML has been enabled for the `pyhtml2text` sub-filter of `html2text` (since it outputs Markdown and AFIK is the only built-in one that does so).
  - Within the filter module the `job.is_markdown` key is set to True when the `pyhtml2text` sub-filter of `html2text` is used.
* HTML output no longer relies on a global stylesheet for formatting since HTML email clients overwrite it (and Gmail even mangles it differently when displaying the content within its webapp vs. displaying it for printing). 
  - Output coloring has been preserved by adding CSS entries to the relevant HTML tags.
  - After these upgrades output is displayed consistenty by the tested HTML email clients and browsers (it wasn't before).
* Wrapping of long lines in the HTML output has been enabled by adding CCS `white-space:pre-wrap` to the HTML pre tag. 
  - This also allows users to set the `line_length: 0` option (do not truncate long lines) of the `pyhtml2text` sub-filter of `html2text`, enabling radically more legible comparisons in most cases.

Minor changes:
* Rewrote footer with correct use of `<address>` tag and making the URL a clickable link.
* Improved speed by replacing `startswith()` with slices (did own testing; see dicussion https://stackoverflow.com/questions/13270888/why-is-startswith-slower-than-slicing)
* Minor Pythonic edit (replaced `localtime=1` with `localtime=True`).

Testing:
* In daily operation since mid-May 

To do:
* Documentation (based on your acceptance of/comments to these changes)

Wish list:
* Improve external library loading and error handling (it's hapazard throughout *urlwatch*)
* Reduce number of mandatory external libraries (do we really need to force `cssselect` on everyone?)
* Figure out a way to do hanging indents (for long lines) using styles that Gmail doesn't break


lib/urlwatch/reporters.py
```diff
@@ -161,37 +161,29 @@ def submit(self):
...
+        yield SafeHtml(
+            '<!DOCTYPE html>\n'
+            '<html>\n'
+            '<head>\n'
+            '<title>urlwatch</title>\n'
+            '<meta http-equiv="content-type" content="text/html; charset=utf-8">\n'
+            '<meta name="viewport" content="width=device-width, initial-scale=1.0">\n'
+            # No stylesheet: styles are embedded to avoid overwriting by email clients
+            '</head>\n'
+            '<body>\n')
```
thp, line 173: Can you please keep the multi-line string here, even if you removed the style sheets?

lib/urlwatch/reporters.py
```diff
@@ -201,49 +193,96 @@ def _parts(self):
...
-            else:
-                yield SafeHtml('<span class="unified_nor">{line}</span>').format(line=line)
+        yield SafeHtml(
+            '<div style="font-family:monospace;font-style:italic">\n'
+            '<span>{pkgname} {version}, {copyright}</span>\n'
+            '<address>Website: <a href="{url}">{url}</a></address>\n'
+            '<span>Watched {count} URLs in {duration} seconds</span>\n'
+            '</div>\n'
+            '</body>\n'
+            '</html>\n').format(pkgname=urlwatch.pkgname, version=urlwatch.__version__, copyright=urlwatch.__copyright__,
```
thp, line 203: Also here, please keep the multi-line string.

lib/urlwatch/reporters.py
```diff
@@ -201,49 +193,96 @@ def _parts(self):
...
+                html_out = htags.sub(r'<\g<1>strong>', html_out)  # replace heading tags with <strong>
+                return pre + html_out
+
+            for line in unified_diff.splitlines():
+                if line[0] == '+':
+                    yield '<span style="color:green">' + '+' + mark_to_html(line[1:]) + '</span>'
+                elif line[0] == '-':
+                    yield '<span style="color:red">' + '-' + mark_to_html(line[1:]) + '</span>'
+                elif line[0] == '@':
+                    yield '<span style="background-color:whitesmoke">' + line + '</span>'
```
thp, line 242: Please use e.g.:

```
'<span style="color:green">+{html}</span>'.format(html=mark_to_html(line[1:]))
```

thp: Also, are you sure this works properly with HTML escaping from the markdown all the time?

mborsetti:
Before
image

After
image

The before is from Gmail, as presented once I clicked on "[Message clipped]  View entire message" to see it.

The after includes #519 but the extra text is immaterial to HTML formatting made by this pull request

lib/urlwatch/reporters.py
thp: Whitespace in HTML doesn't matter. If it is a problem use [`textwrap.dedent()`](https://docs.python.org/3/library/textwrap.html#textwrap.dedent) on the multi-line string.

lib/urlwatch/reporters.py
thp: Also here, use [`textwrap.dedent()`](https://docs.python.org/3/library/textwrap.html#textwrap.dedent) if it is an issue. Multi-line strings are better/less noisy than this (less quotes, less `\n`).

lib/urlwatch/reporters.py
thp: `str.format()` and later literal string interpolation (aka f-strings, since Python 3.6) are better to read and maintain. Less `'` and `+` and if the string to be inserted is a big expression, the expression is named and the place where it is inserted is easier to see (e.g. `{html}` instead of `' + mark_to_html(line[1:])) + '`).

Also it's easier to make it customizable -- you could let the user provide a string and document which keys are provided without having to modify the code, and the user can move the `{html}` around or even use it multiple times (e.g. `<strong title="{html}">Added line: {html}</strong>`). Not that we should make it customizable, but using strings + `.format()` makes this super easy also in other parts of the code.

lib/urlwatch/reporters.py
```diff
@@ -161,37 +161,29 @@ def submit(self):
...
+                title = '<span title="{location}" style="color:#888888">{pretty_name}</span>'
             else:
                 title = '{location}'
-            title = '<h2><span class="verb">{verb}:</span> ' + title + '</h2>'
+            title = '<h2><span style="color:#888888">{verb}:</span> ' + title + '</h2>'
 
-            yield SafeHtml(title).format(verb=job_state.verb,
+            yield SafeHtml(title).format(verb=job_state.verb.title(),
                                          location=job.get_location(),
                                          pretty_name=job.pretty_name())
```
thp, line 188: In fact, here's an example how we use `str.format()` to separate the string contents from the formatting step (related to the question below regarding why `str.format()` is better than string concatenation a'la `' + ... + '`).

lib/urlwatch/reporters.py
mborsetti: Thanks for taking the time for the explanation. 

I do personally use f-strings and had to revert to using '+' (now `str.format()`) to testing for 3.5.

mborsetti:
Made all changes requested and rebased.

lib/urlwatch/filters.py
```diff
@@ -260,7 +260,8 @@ def filter(self, data, subfilter):
         else:
             method = 're'
             options = {}
-
+        if method == 'pyhtml2text':
+            setattr(self.job, 'is_markdown', True)
```
thp, line 264: This can just be:

    self.job.is_markdown = True

No need to use `setattr()` here.

lib/urlwatch/reporters.py
```diff
@@ -175,48 +166,98 @@ def _parts(self):
...
+            # but Gmail strips the 'text-indent:-2.3em' and end up with everything indented
             return ''.join((
-                '<pre>',
+                '<pre style="white-space:pre-wrap">',
                 '\n'.join(self._diff_to_html(job_state.get_diff())),
                 '</pre>',
             ))
+        elif difftype == 'table':
+            timestamp_old = email.utils.formatdate(job_state.timestamp, localtime=1)
+            timestamp_new = email.utils.formatdate(time.time(), localtime=1)
```
thp, line 248: `localtime=True`

lib/urlwatch/reporters.py
```diff
@@ -175,48 +166,98 @@ def _parts(self):
...
+            <div style="font-family:monospace;font-style:italic">
+            <span>{pkgname} {version}, {copyright}</span>
+            <address>Website: <a href="{url}">{url}</a></address>
+            <span>Watched {count} URLs in {duration} seconds</span>
+            </div>
+            </body>
+            </html>""").format(pkgname=urlwatch.pkgname, version=urlwatch.__version__, copyright=urlwatch.__copyright__,
+                               url=urlwatch.__url__, count=len(self.job_states), duration=self.duration.seconds)
+
+    def _diff_to_html(self, unified_diff, is_markdown=False):
```
thp, line 178: Since this function isn't using the `self` object, it could be made a stateless global function (`def _diff_to_html(unified_diff, is_markdown=False);` in module-global scope) and have some basic unit tests added.

lib/urlwatch/reporters.py
```diff
@@ -175,48 +166,98 @@ def _parts(self):
...
+            htags = re.compile(r'<(/?)h\d>')
+            tags = re.compile(r'^<p>(<code>)?|(</code>)?</p>$')
+
+            def mark_to_html(text):
+                '''converts Markdown (as generated by pyhtml2text filter) back to html'''
+                if text == '* * *':  # manually expand horizontal ruler since <hr> tag is used by us to separate jobs
+                    return '-' * 80
+                pre = ''
+                if text.lstrip().startswith('* '):  # item of unordered list
+                    lstripped = text.lstrip()
```
thp, line 192: `str.lstrip()` by default removes all whitespace, if it's a mix of tabs and spaces, the indentation will likely be wrong, but it's probably fine, just wanted to point it out.

lib/urlwatch/reporters.py
```diff
@@ -175,48 +166,98 @@ def _parts(self):
...
+
+    def _diff_to_html(self, unified_diff, is_markdown=False):
+        if is_markdown and Markdown:
+            # rebuild html from markdown using markdonw2 library's Markdown
+            markdowner = Markdown(safe_mode='escape', extras=['strike', 'target-blank-links'])
+            atags = re.compile(r'<a ')
+            htags = re.compile(r'<(/?)h\d>')
+            tags = re.compile(r'^<p>(<code>)?|(</code>)?</p>$')
+
+            def mark_to_html(text):
```
thp, line 186: This inline-defined function could also be moved to global scope and have some unit tests or at least doctests to test for the most common replacements.

lib/urlwatch/reporters.py
```diff
@@ -175,48 +166,98 @@ def _parts(self):
...
+            </html>""").format(pkgname=urlwatch.pkgname, version=urlwatch.__version__, copyright=urlwatch.__copyright__,
+                               url=urlwatch.__url__, count=len(self.job_states), duration=self.duration.seconds)
+
+    def _diff_to_html(self, unified_diff, is_markdown=False):
+        if is_markdown and Markdown:
+            # rebuild html from markdown using markdonw2 library's Markdown
+            markdowner = Markdown(safe_mode='escape', extras=['strike', 'target-blank-links'])
+            atags = re.compile(r'<a ')
+            htags = re.compile(r'<(/?)h\d>')
+            tags = re.compile(r'^<p>(<code>)?|(</code>)?</p>$')
```
thp, line 184: Since `atags`, `htags` and `tags` are only used once below, there's no benefit of using `re.compile`, can just use the `re.sub` / `re.subn` functions directly.

lib/urlwatch/reporters.py
mborsetti: Sorry, don't have the time to refactor existing code.  This contribution is already costing me way too much,

lib/urlwatch/reporters.py
mborsetti: Missed that upgrade to legacy code; done.

lib/urlwatch/reporters.py
mborsetti: Same as above on refactoring; however, I did build unit tests.

lib/urlwatch/reporters.py
mborsetti: Are you sure about that?  They're used by the function `mark_to_html(text)`, which is called on (almost) every single line.

lib/urlwatch/reporters.py
thp: > Sorry, don't have the time to refactor existing code. This contribution is already costing me way too much,

Well I guess then I am not going to spend my time to review your PR. Feel free to maintain this thing in your fork then.

mborsetti: Will do, thanks.

Issue #515, opened 2020-06-27
Feature Request - Diff to Script


Shaav:
Forgive me if there's already a way to accomplish this that I overlooked—I don't think that it's possible right now...

I was thinking it would be really useful to have the ability to pass the diff output of a job to custom script. The specific use-case I had in mind is a webpage I monitor for additional links. I would like to pull the new links added and then trigger wget on them. What I'm imagining is pretty simple I think—something like:

    name: "My Job"
    url:https://example.com/links.html
    filter: element-by-tag:a
    script: /<path-to-script>/process_new_links.sh

I guess it's kind of like a custom reporter, but I wouldn't want to run all my jobs through it.  I think it would also be useful to optionally have the output of the script be passed back to the reporter or to ignore it and just report the diff like usual.

thp:
Why not use the `stdout` reporter and pipe urlwatch output into your custom script? You can have a separate configuration and jobs file for the job you want to do this for and keep the rest of the configuration separate.

Shaav:
> Why not use the `stdout` reporter and pipe urlwatch output into your custom script? You can have a separate configuration and jobs file for the job you want to do this for and keep the rest of the configuration separate.

Thanks for the suggestion, I'll give that a try.

It's actually 10 or 12 different jobs so that becomes a lot of overhead keeping and calling that many different configuration files. Or I guess parsing through the output based on the titles.

thp:
As an alternative, you can set `diff_tool` on the job and it gets the old a new content as filename parameters (the files will be temporary files that contain the old and new page content) and you can do whatever you want there as a side effect.

thp:
In the Git master branch (not yet in any release) it's now possible to specify a `diff_filter` that can - in combination with `shellpipe` be used for that. As an added bonus, the output of the script is the thing that urlwatch reports as the diff contents, so if the script outputs the results of processing, you can use urlwatch to send the processing results to any of the supported reporters.

This is now also documented as "Advanced Topic" here:
https://urlwatch.readthedocs.io/en/latest/advanced.html#pass-diff-output-to-a-custom-script

Shaav:
Awesome! You are a scholar and a gentleman!  Thanks! :)

Issue #407, opened 2019-05-30
get notified of new facebook events


fabienmorin:
I don't have any FB account but I'm trying to get informed of some public pages. Especially, I'd like to get notified when there is a new event.

For this, I tried the following config :
kind: url
name: Facebook test
url: https://www.facebook.com/pg/FacebookPageIWantToFollow/events/?ref=page_internal
filter:
    - element-by-id: upcoming_events_card
    - html2text: pyhtml2text

but it doesn't work as expected, I never get notified, even when a new event is added whereas it works for other sites than FB. Any idea how I could achieve that ?

thp:
Does the page require to be logged in? What if you add the cookies from your browser so the settings so that it can use your logged-in session?

fabienmorin:
no, as I said before, I have no FB account but I'd liked to watch public pages to get notified in case of new event (which is possible to do by hand without account)

mborsetti:
Have you tried debugging using `--test filter` as per docs (https://urlwatch.readthedocs.io/en/latest/filters.html)?

Also, when I go to that page from an Incognito browser, that page looks like below, and I presume it rarely changes (so there's very little for urlwatch to pick up).  If that's not what you're trying to monitor, then you're using an incorrect URL.

![image](https://user-images.githubusercontent.com/26698111/87505943-be696f00-c69c-11ea-9658-2ae6049043cf.png)

thp:
> no, as I said before, I have no FB account but I'd liked to watch public pages to get notified in case of new event (which is possible to do by hand without account)

I have given it a try and it seems Facebook loads data dynamically, which is why you don't see the events. As a workaround, I tried the following and this worked for me (you can test with `--test-filter`, obviously you need to replace `FacebookPageIWantToFollow` with a real Facebook page that has events):

```
navigate: https://m.facebook.com/FacebookPageIWantToFollow/pages/permalink/?view_type=tab_events
filter:
  - css: div._5zma
  - html2text
```

Some things to note:

1. Use `m.facebook.com`, which is a simpler page format
2. Use `navigate` to use [requests-HTML](https://requests.readthedocs.io/projects/requests-html/en/latest/) for executing JavaScript (this requires `python3 -m pip install requests_html`)
3. On the mobile page, for the "Event" click on "See All" (this is `/pages/permalink/?view_type=tab_events`)
[...]

PR #519, opened 2020-07-06
Filter unified diff to keep only addition or deleted lines

mborsetti:
Filter unified diff to keep only addition or deleted lines

The report from the unified diff can now be filtered to keep only addition lines or deleted lines.
 - Added job key `comparison_filter`, which can take the values of `additions` or `deletions`.
 - A value of `additions` will cause reports to contain only lines that are added by the diff (no deletions).
 - A value of `deleted` key will cause reports to contain only lines that are deleted by the diff (no additions). 
`comparison_filter: additions` is extremely useful for monitoring new content on sites where content gets added while old content "scrolls" away.

Because lines that are modified generate both a deleted and an added line by the diff, this filters always displays modified lines.

As a safeguard, `additions` will warn when 75% of more of the change consists of deletions.

Sample output for `additions`:
```
CHANGED: https://example.com
... @   Sat, 23 May 2020 00:00:00 +0000
+++ @   Sat, 23 May 2020 01:00:00 +0000
-**Comparison type: Additions only**
@@ -1,2 +1,2 @@
+This is a line that has been added or changed
```

Sample output for `deletions`:
```
CHANGED: https://example.com
--- @   Sat, 23 May 2020 00:00:00 +0000
... @   Sat, 23 May 2020 01:00:00 +0000
+**Comparison type: Deletions only**
@@ -1,2 +1,2 @@
-This is a line that has been deleted or changed
```

Sample output for `additions` when the 75% deletions safeguard is triggered:
```
CHANGED: https://example.com
... @   Sat, 23 May 2020 00:00:00 +0000
+++ @   Sat, 23 May 2020 01:00:00 +0000
-**Comparison type: Additions only**
.** No additions (only deletions)
--- WARNING: 39 lines deleted; suggest checking source
```

Bug:
* Due to legacy logic in urlwatch, I am unable to suppress reporting when all changes are filtered out, so a message is added instead:
```
CHANGED: https://example.com
... @   Sat, 23 May 2020 00:00:00 +0000
+++ @   Sat, 23 May 2020 01:00:00 +0000
-**Comparison type: Additions only**
.** No additions (only deletions)
```
Would love some pointers on how to implement logic to suppress reporting instead.

Testing:
* In daily operation since mid-May

To do:
* Documentation (based on your acceptance of/comments to these changes)


mborsetti:
> I wonder if it would make more sense to move this logic into an external `diff_tool` (which in turn could call a normal "diff" and then do the filtering via some scripting). 

The `diff_tool` key calls `subprocess.Popen` making it a platform-dependent solution. This is sub-optimal and personally does not work as I have to run `urwatch` on both linux and Windows machines.

> This way, there would be no need to add this feature into the urlwatch core, but we could document a script to be used as `diff_tool` in the documentation and this way be more flexible (e.g. `comparison_filter` currently doesn't work at all in combination with `diff_tool`, but users might expect it to).

I do not see how there could be an expectation for `comparison_filter` to work in combination with `diff_tool` given that the latter can return any arbitrary text. Maybe it's my choice of name for the key that suggests this? Should it be named e.g. `unified_diff_comparison` or `unified_diff_filter` instead? Ideas/suggestions?

> Suppression of output could be implement (if it doesn't already work that way) by returning an empty string from `unified_diff()` -- I haven't checked if that requires lots of refactoring.

The detection of whether anything has changed is made a priori of what the differ returns, so refactoring would indeed be required. But I easily get lost in the code and its various classes/nestings/function calls so any help would be appreciated.  This could be a future upgrade.

> The `localtime=1` -> `localtime=True` change is good, we could do this in a separate cleanup commit, anyway.

Guilty as charged on being sloppy with `localtime=1`; will separate it into its own pull request.

thp:
`subprocess.Popen()` should work just fine on Windows. Also these days, with WSL you can get a Linux environment set up in Windows no problem.

docs/source/jobs.rst
```diff
@@ -109,11 +109,83 @@ Job-specific optional keys:
...
+   ---------------------------------------------------------------------------
+   ... @   Sat, 23 May 2020 00:00:00 +0000
+   +++ @   Sat, 23 May 2020 01:00:00 +0000
+   -**Comparison type: Additions only**
+   .** No additions (only deletions)
+   --- WARNING: 39 lines deleted; suggest checking source
+   ---------------------------------------------------------------------------
+
+
+Workaround: due to legacy logic in urlwatch, we are unable to suppress reporting when all changes are filtered out, so a message is added instead:
```
thp, line 178: It would be great if we could just fix the legacy logic so that instead of documenting the workaround we'll fix it properly. A quick look at `reporters.py` shows that each job state gets added to the summary and details like this:

```
        summary = []                                                                                                     
        details = []                                                                                                     
        for job_state in self.report.get_filtered_job_states(self.job_states):                                           
            summary_part, details_part = self._format_output(job_state)                                                  
            summary.extend(summary_part)                                                                                 
            details.extend(details_part)                                                                                 
```

And `self._format_output()` there calls `self._format_content()` which in turn calls `self.unified_diff()`, so it would be possible that if `self.unified_diff()` returns an empty string (or `None`), then `_format_output()` handles that and the return value of that function is handled on the caller site. The only tricky thing is that the "minimal" reporter doesn't call the diffing at all, so with the minimal reporter, changes (as summary line) would always be displayed, but that's likely fine, as it doesn't affect the diff output then, anyway.

docs/source/jobs.rst
```diff
@@ -109,11 +109,83 @@ Job-specific optional keys:
 
 Optional keys for all job types
 -------------------------------
-
+- :ref:`comparison_filter`: fiter unified diff output to keep only addition lines or deleted lines
```
thp, line 112: Typo: `fiter`

docs/source/jobs.rst
```diff
@@ -109,11 +109,83 @@ Job-specific optional keys:
...
+   ---------------------------------------------------------------------------
+   CHANGED: https://example.com
+   ---------------------------------------------------------------------------
+   --- @   Sat, 23 May 2020 00:00:00 +0000
+   ... @   Sat, 23 May 2020 01:00:00 +0000
+   +**Comparison type: Deletions only**
+   @@ -1,2 +1,2 @@
+   -This is a line that has been deleted or changed
+
+Sample output for `additions` when the 75% deletions safeguard is triggered:
```
thp, line 163: Would it be better to just show the deletions in this case?

lib/urlwatch/reporters.py
```diff
@@ -76,6 +76,35 @@
 WDIFF_REMOVED_RE = r'[\[][-].*?[-][]]'
 
 
+@functools.lru_cache(maxsize=1)
+def cached_unified_diff(old_data, new_data, timestamp_old, timestamp_new, comparison_filter):
```
thp, line 80: This could use a unit test or three :)

docs/source/jobs.rst
mborsetti: Great idea! 

Done in 7af11c4

docs/source/jobs.rst
mborsetti: Great eye!

Done as part of 7af11c4

docs/source/jobs.rst
mborsetti: Thanks for the pointers.  Will take a look at it.  I am still overly confused at the Russian dolls here... :)

lib/urlwatch/reporters.py
mborsetti: Hmm... but there are no errors if we don't test! ;)

OK great idea.

mborsetti: OK, I took a look at the (undocumented) files in the `test` folder and have absolutely no idea on how to do unit testing of this job filter. I believe it would require running a job with a specified filter to set the baseline, then running another job on the same source with a different filter that has been modified to generate a diff (e.g. through the use of re.sub or grep/grepi), capture its output, and see if it matches expectations.  

I couldn't find anything in there remotely resembling this and I cannot understand enough about the structure of the program the nesting inside a nest inside a nest to write one from scratch -- nor do I have the time to do so.

I am afraid it's something someone else will have to tackle.

thp: The `cached_unified_diff()` doesn't have anything to do with jobs.

Create a function `test_cache_unified_diff()` in a new file `test_reporters.py` and add something like this:

```python
from urlwatch.reporters import cache_unified_diff
from nose.tools import eq_

def test_cache_unified_diff_normal():
    input_text_a = 'a'
    input_text_b = 'b'
    expected_diff_output = "..."  # put here the result you expect the function to return
    eq_(cache_unified_diff(input_text_a, input_text_b, '', '', None), expected_diff_output)
```

And then add tests for other cases, e.g.:

- changed line with "additions" filter
- changed line with "deletions" filter
- "additions" filter with more than 75% deletions
- "additions" filter and lines were only deleted
- "deletions" filter and lines were only added

thp:
I still think this might be better suited as an external diff tool, which might be as simple as a shell script (e.g. for showing only additions):

    diff -u "$@" | grep '^+' | grep -v '^+++'

This way, people also have a template for how to easily customize their diffs. Something that would be very good for the core is to add the ability to avoid reporting if the diff tool results in empty output.

Of course, it might make sense to make it a Python script that uses the code you have here. It just feels like we're adding bloat to core parts of the codebase that could be done with external tools.

lib/urlwatch/reporters.py
mborsetti: Thanks!!! Unit testing written

docs/source/jobs.rst
mborsetti: Thanks for the refactoring, I got it to work!

mborsetti:
This filter should be added to the recipe for #407 since the person only wanted to be notified of "new events", not when old events scroll out of the page they're monitoring.

thp:
Looking again at #515 I wonder if we should just allow `diff_filter` as key and it's the same filter list as the `filter` key and treated the same, just that instead of filtering the old and new contents before the diffing, it will take the diff result and run the filters on those and then use that as the change result.

This would allow passing the diff results to a script as side effect (using `shellpipe`) as well as allow writing of the filters you do to filter out additions and deletions, plus more flexible things (e.g. filter only additions and then grep for a specific word, so you can even be notified only when a certain word is added to the page for example).

PR #524, opened 2020-07-11
Upgrade in docs

mborsetti:
Significant upgrade in documentation.  Not perfect yet, but want to start pushing it to coordinate with #522 (e.g. this update already documents filters the way you want it) and other work being done in the documentation.


mborsetti:
> It would be great if you could split down this big PR into smaller documentation updates, this way it's easier to quickly merge things that obviously can be merged and discuss other things that might need changing.

This will be impossible to redo from scratch into chunks as it will take hours on ends; sorry.  

> A lot has changed in the filters documentation in #522, I have also added unit tests for all filter examples (have a look at the PR -- using unique URLs in the examples and mapping them to expected input/output and then using docutils to extract the YAML code blocks from the filters documentation).

If you take a look at it you will see that quite a lot of work has been done, including on (finally) providing documentation for every single built-in filters, and it's not as simple as wordsmithing it.  I love what you did in #522 (I was unaware that https://www.example.org/ existed!) and can manually fold your changes into this to bring it up to date and unify it; I know it's going to be a time-consuming pain but am prepared to do it if it's going to be used.

docs/source/jobs.rst
```diff
@@ -53,35 +57,42 @@ Job-specific optional keys:
...
-
-The optional ``requests-html`` package must be installed to run "Navigate" jobs
-(see :ref:`dependencies`).
-
-.. code-block:: yaml
-
+The Navigate jobs are currently not working.
+
+They are intended to handle web pages requiring JavaScript in order to render
+the content to be monitored.
```
thp, line 63: As far as I could tell from my experiments in #407, navigate jobs are in fact working (at least for that use case there).

thp:
> If you take a look at it you will see that quite a lot of work has been done, including on (finally) providing documentation for every single built-in filters, and it's not as simple as wordsmithing it. I love what you did in #522 (I was unaware that https://www.example.org/ existed!) and can manually fold your changes into this to bring it up to date and unify it; I know it's going to be a time-consuming pain but am prepared to do it if it's going to be used.

I can't merge it as-is, sorry. I also can't promise that I'll merge anything you submit. The timing of this with #522 wasn't ideal, but it is what it is. In general, submitting smaller PRs increases the chances of getting quick reviews and merges. Of course that doesn't apply to me as the maintainer, I'm well aware that #522 was quite a big chunk, but sometimes it's needed to push things forward. In general, I also try to keep the PRs I do as small as possible.

Having documentation (and now since #522 also implicit testing of the documented examples) of all filters is of course very welcome, some of the other changes in this PR I feel like we shouldn't merge, at least not as-is.

docs/source/jobs.rst
cfbao: I just checked. That case worked because Facebook loads events statically without using JavaScript on mobile pages. There's still no browser or JavaScript running with navigate/browser jobs. It's the same as using url jobs.

thp: The mobile Facebook page seems to indeed also contain the content, but there's still a different HTML content being loaded by `navigate`, so it does "something" (the CSS selectors in #407 don't work on the non-javascript page).

thp:
Please submit documentation updates in smaller PRs. Closing this for now, as it can't be merged as-is, anyway.

Issue #452, opened 2020-02-22
Grep options?


jprokos:
It would be helpful to use grep options:

`-i, --ignore-case`

Is this possible? Can we set this in the config file somehow?

If this is already possible would someone provide a working example or notate it in the man page?

cfbao:
No grep command line options are supported. Urlwatch doesn't actually use the `grep` utility internally.

However, you can use the inline flag `(?i)` to turn on case insensitive mode.
https://www.regular-expressions.info/modifiers.html#bodytext
https://docs.python.org/3/library/re.html#re.I

jprokos:
Thanks, this is exactly what I needed.
Please provide an example of using the inline flag in a urlwatch filter. I am not sure of the placement.

jprokos:
Thanks, this is exactly what I needed.
Here is an example that works for me:

`- grep: (?i)olympic`

jprokos:
Thank you so much, this is exactly what I needed. I found the answers to my other questions using your first link. I couldn't find that information. 

Here is an example that works for me:

`- grep: (?i)olympic`

And when searching for more than one string the first inline flag affects the entire line.

`- grep: (?i)olympic|time standards|2020`
The case will stay the same for the entire expression.

If you want to specify case sensitivity in the middle of the expression you can turn off case insensitive mode using the minus sign:

`- grep: (?i)olympic|(?-i)Time Standard|2020`

jprokos:
I am only able to get the (?i) flag to work. Other flags fail... e,F

Is there something I am missing or not understanding?

jprokos:
Was not able to use the (?v) flag either to exclude items from grep search.

jprokos:
I have this rule. I am trying to use grep to remove an item from results but I am having difficulty.


```name: MacRumors Apple Releases
kind: url
url: "https://www.macrumors.com"
filter:
 - xpath: '(//div[contains(@class,"wrapper")]//h2)/a'
 - grep: (?i)Apple Seeds|Apple Releases
 - html2text: re
---
```

These are the results:

```$ uwtf 22
Apple Seeds Fifth Beta of watchOS 6.2 to Developers
Apple Seeds Fifth Betas of iOS and iPadOS 13.4 to Developers [Update: Public Beta Available]
Apple Seeds Fifth Beta of Upcoming macOS Catalina 10.15.4 Update to Developers [Update: Public Beta Available]
[...]

Issue #458, opened 2020-03-01
Possible to get the imediate lines before or after grep pattern?


DazzyWalkman:
Just as what shell grep -A or -B do.
My url source contains blocks of lines like 

> \<p>datetime<\/p>
> \<p>title<\/p>

I grep for a specific title, that have multiple occurrences on that page. Grep only returns the title lines,  that defeats my purpose of update checking.
So I also need the respective datetime line.
Thank you.

jprokos:
You can grep for more than one line by using the pipe.

Example:
`'(//div[contains(@class,"item")]//h3)/a|(//div[contains(@class,"item")]//h4)/a'`

thp:
In the master branch (not yet in a release) there's now the `shellpipe` filter which you can use to pipe the content through a shell command, so you can use the `grep` command line utility with any parameters you want.

This is documented with examples here:
https://urlwatch.readthedocs.io/en/latest/filters.html#using-a-shell-script-as-a-filter

thp:
This is now in urlwatch 2.19.

Issue #419, opened 2019-09-10
Grep filter assist


strauhmanis:
Hi!
This is a request for assist with grep filter, if you could, please.

The goal is to check/store the price, as follows:

`url: "https://apps.apple.com/us/app/plantsnap-plant-identification/id1234702272"`
`filter: grep:(@type":"Offer"\054"price":)\d+[\054.]?\d+`

Viewing the `cache.db` after urlwatch processed the page, I see that the the whole line, of which the above section `@type":"Offer","price":9.99` is part of, is being stored. And the line is some 7000+ characters.

Please advise how do I make urlwatch to store only the `@type":"Offer","price":9.99` part.

Many thanks in advance!

mcint:
Not currently possible that way.

Maybe use the RegexFilter with `(.*)(your-pattern)(.*)` and return capture group 2. It might possible, but it will have terrible performance. Although, it looks like it can only replace with a predetermined fixed pattern :(.

If this shelled out to grep, `-o` but [current behavior](https://github.com/thp/urlwatch/blob/67b7146e20dc2c4fae8650fcd726ef46f0f9b737/lib/urlwatch/filters.py#L202-L212) is `line ... if re.search(subfilter, line)`.

You could submit a patch. I do wish urlwatch played nicer with existing tools, instead of half implementing features.

thp:
In the master branch (not yet in a release) there's now the `shellpipe` filter which you can use to pipe the content through a shell command, so you can use the `grep` command line utility with any parameters you want.

This is documented with examples here (the first example is probably the use case you want):
https://urlwatch.readthedocs.io/en/latest/filters.html#using-a-shell-script-as-a-filter

thp:
This is now in urlwatch 2.19.

PR #525, opened 2020-07-11
Add new "shellpipe" filter for filtering via command-line utilities

thp:
For #452 and others.


thp:
This should also help with #419.

thp:
And also with #458.

PR #550, opened 2020-08-12
Fix diff generation for lines without newline

ad1217:
Rather than keeping the line endings before passing to diff, then joining the resulting lines with '', strip the line endings and join
with '\n'. This ensures a that each output line has exactly one trailing newline.

Possible issues:
 - This will prevent detection of a difference in line endings.
 - Joining with `os.linesep` might be better for Windows, but only if it is going to a text file, and I think most Windows stuff handles lines terminated with just LF fine now.

### DIFF
--- a/lib/urlwatch/handler.py
+++ b/lib/urlwatch/handler.py
@@ -161,9 +161,9 @@ class JobState(object):
 
         timestamp_old = email.utils.formatdate(self.timestamp, localtime=True)
         timestamp_new = email.utils.formatdate(time.time(), localtime=True)
-        return ''.join(difflib.unified_diff(self.old_data.splitlines(keepends=True),
-                                            self.new_data.splitlines(keepends=True),
-                                            '@', '@', timestamp_old, timestamp_new))
+        return '\n'.join(difflib.unified_diff(self.old_data.splitlines(),
+                                              self.new_data.splitlines(),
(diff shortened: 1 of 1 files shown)


polyzen:
Before:
```
@@ -1,3 +1,3 @@
-1.5.8
-2020-07-01T04:51:11
-707ed1bdffd6b8bec2d74dca36cc2665081c8db8590241bf6a72a3378928dd65+1.5.9
+2020-07-23T04:58:02
+6f3e18befd0f26118c6c6ddf39e2ef18366ff9501cbcbf664ffde669e3c2d132
```

After:
```
@@ -1,3 +1,3 @@
-1.5.8
-2020-07-01T04:51:11
-707ed1bdffd6b8bec2d74dca36cc2665081c8db8590241bf6a72a3378928dd65
+1.5.9
+2020-07-23T04:58:02
+6f3e18befd0f26118c6c6ddf39e2ef18366ff9501cbcbf664ffde669e3c2d132
```

With this filter:
https://github.com/polyzen/dotfiles/blob/3b83c549d2e497098564ad7fdbb44c5050379231/misc/.config/urlwatch/hooks.py#L52

thp:
This is a good change, I've ran into the issue as described by @polyzen already a few times, good that this is now fixed. I think most people don't care about newline changes, and if they do, they can always use an external `diff_tool` or the `hexdump` filter.

Change documented here: c3b5c24

PR #553, opened 2020-08-13
WIP: Add RSS Reporter

### LINKED ISSUE #53: Can urlwatch generate RSS feeds?
I was wondering If I could get the differences of the webpage in RSS format. This would result in having a custom hosted alternative to sites like [Feed43](http://feed43.com/), [Page2RSS](http://page2rss.com/) etc.

ad1217:
This is a work in progress attempt at a RSS reporter. It retrieves job history and compares each version with the previous one to generate a hopefully stable RSS feed.

I have not tested this very thoroughly (it has no tests, but neither do the other reporters) and it is missing documentation, among other things.

I suspect this will either break or be able to be improved once #360 (or equivalent) is merged.

Resolves #53.

### DIFF
--- a/lib/urlwatch/handler.py
+++ b/lib/urlwatch/handler.py
@@ -133,16 +133,16 @@ class JobState(object):
 
         return self
 
-    def get_diff(self):
+    def get_diff(self, new_time=None):
         if self._generated_diff is None:
-            self._generated_diff = self._generate_diff()
+            self._generated_diff = self._generate_diff(new_time)
             # Apply any specified diff filters
             for filter_kind, subfilter in FilterBase.normalize_filter_list(self.job.diff_filter):
                 self._generated_diff = FilterBase.process(filter_kind, subfilter, self, self._generated_diff)
 
         return self._generated_diff
 
-    def _generate_diff(self):
+    def _generate_diff(self, new_time=None):
         if self.job.diff_tool is not None:
             with tempfile.TemporaryDirectory() as tmpdir:
                 old_file_path = os.path.join(tmpdir, 'old_file')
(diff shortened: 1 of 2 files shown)


thp:
Would the logic used for `--test-diff-filter` help there?

ad1217:
Yup, that's pretty much exactly the same logic. I'm not sure where a method containing that logic could be placed nicely though; it doesn't seem to fit well in either `JobBase` or `JobState`.

thp:
You could make it just another command-line argument instead of a fully-fledged reporter and then you can put it in `lib/urlwatch/command.py`. A reporter usually only runs when there's a change, but in this case, I could imagine people wanting to export an RSS at some point where everything is already set up (of course, these days, one could use `--test-reporter`, but that's basically working around the fact that it's not really a reporter, but something meant to be run as command). There can, however, also be a "reporter" in the sense that it would automatically redo the RSS export. So feel free to even create a `lib/urlwatch/rss.py` and put things there (it's probably fine to use a function from there in the implementation of `--test-diff-filter`, otherwise put the shared function in e.g. `lib/urlwatch/history.py` or something).

I would imagine this feature working something like this:

```
# Export an RSS file, one-off command that exports the full history
urlwatch --export-rss /path/to/something.rss
```

Plus a reporter that takes as configuration argument at least the `filename` and will - when enabled - do the RSS export every time a change is detected, basically automating the `--export-rss` for cases where one has urlwatch running on a server.

In the future (but this can be scoped out of the PR to keep it simple) i'm sure people will want to create separate RSS files for each job of for a subset of jobs, etc... -- this should be easy to add once the basic infrastructure is in, but no need to do it in this PR, imho.

thp:
Closing due to inactivity.

Issue #353, opened 2019-01-25
Make MAX_LENGTH in SlackReporter configurable


nscaife:
Some Slack-compatible webhooks (i.e., [Discord](https://discordapp.com/developers/docs/resources/webhook#execute-webhook-jsonform-params)) have different maximum message lengths.  Discord has a maximum message length of 2,000 characters for webhooks.

I verified that Discord rejects >2,000 character messages from urlwatch and that manually adjusting `MAX_LENGTH = 2000` in SlackReporter corrects the issue.  Can this be made into a configurable value for this reporter?

thp:
Please create a pull request with the proposed change.

PR #566, opened 2020-09-11
Enable configuring max message length for Slack-compatible webhooks

YusefOuda:
Addresses https://github.com/thp/urlwatch/issues/353

### DIFF
--- a/lib/urlwatch/reporters.py
+++ b/lib/urlwatch/reporters.py
@@ -610,9 +610,12 @@ class TelegramReporter(TextReporter):
 
 class SlackReporter(TextReporter):
     """Send a message to a Slack channel"""
-    MAX_LENGTH = 40000
 
     __kind__ = 'slack'
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.MAX_LENGTH = self.config['max_message_length'] or 40000
 
     def submit(self):
         webhook_url = self.config['webhook_url']
--- a/lib/urlwatch/reporters.py
+++ b/lib/urlwatch/reporters.py
@@ -615,7 +615,7 @@ class SlackReporter(TextReporter):
 
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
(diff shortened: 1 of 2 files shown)


lib/urlwatch/storage.py
```diff
@@ -124,6 +124,7 @@
         'slack': {
             'enabled': False,
             'webhook_url': '',
+            'max_message_length': 40000
```
thp, line 127: Add trailing comma here

lib/urlwatch/reporters.py
```diff
@@ -610,10 +610,13 @@ def submitToTelegram(self, bot_token, chat_id, text):
 
 class SlackReporter(TextReporter):
     """Send a message to a Slack channel"""
-    MAX_LENGTH = 40000
 
     __kind__ = 'slack'
 
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.MAX_LENGTH = self.config['max_message_length'] or 40000
```
thp, line 618: Since it might not be set, use `self.config.get('max_message_length', 40000),

thp (changes requested):
See comments

lib/urlwatch/reporters.py
```diff
@@ -610,10 +610,13 @@ def submitToTelegram(self, bot_token, chat_id, text):
 
 class SlackReporter(TextReporter):
     """Send a message to a Slack channel"""
-    MAX_LENGTH = 40000
 
     __kind__ = 'slack'
 
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.MAX_LENGTH = self.config.get('max_message_length', 40000)
```
thp, line 618: Sorry to nitpick again, but since it's not a constant anymore, prefer lowercase:

    self.max_length = ...

This also means changing all other occurences of `self.MAX_LENGTH` to `self.max_length`.

After that, it's ready to merge.

thp (changes requested):
One small nitpick, then it's ready to merge.

thp (approved):

thp:
Thanks!

PR #572, opened 2020-09-17
Some Matrix reporter improvements.

dkasak:
This includes:

- Smarter logic for report trimming (to meet the maximum length requirement) so
  we don't end up with invalid Markdown.
- Proper rendering when a code block starts with empty lines.
- Preparation for syntax highlighting for diffs (pending on a resolution of an
  [upstream issue](https://github.com/trentm/python-markdown2/issues/370)).
- Increase in message max length.

### DIFF
--- a/lib/urlwatch/reporters.py
+++ b/lib/urlwatch/reporters.py
@@ -731,12 +731,12 @@ class MarkdownReporter(ReporterBase):
         trim_message_length = len(trim_message)
 
         if max_length is None or len(s) + wrapper_length <= max_length:
-            return "```\n{}\n```".format(s)
+            return "```diff\n{}\n```".format(s)
         else:
             while len(s) + wrapper_length + trim_message_length > max_length:
                 s = re.sub("\n.*$", "", s)
 
-            return "```\n{}\n```\n{}".format(s, trim_message)
+            return "```diff\n{}\n```\n{}".format(s, trim_message)
 
     def _format_content(self, job_state):
         if job_state.verb == 'error':
@@ -794,7 +794,7 @@ class MatrixReporter(MarkdownReporter):
         client_api = matrix_client.api.MatrixHttpApi(homeserver_url, access_token)
 
         if Markdown is not None:
-            body_html = Markdown(extras=["fenced-code-blocks"]).convert(body_markdown)
+            body_html = Markdown(extras=["fenced-code-blocks", "highlightjs-lang"]).convert(body_markdown)
 
             client_api.send_message_event(
                 room_id,
--- a/lib/urlwatch/reporters.py
+++ b/lib/urlwatch/reporters.py
@@ -742,7 +742,7 @@ class MatrixReporter(MarkdownReporter):
         client_api = matrix_client.api.MatrixHttpApi(homeserver_url, access_token)
 
         if Markdown is not None:
-            body_html = Markdown().convert(body_markdown)
+            body_html = Markdown(extras=["fenced-code-blocks"]).convert(body_markdown)
 
             client_api.send_message_event(
                 room_id,
--- a/lib/urlwatch/reporters.py
+++ b/lib/urlwatch/reporters.py
@@ -668,88 +668,144 @@ class MarkdownReporter(ReporterBase):
             summary.extend(summary_part)
             details.extend(details_part)
 
-        # Sum of the lengths of parts, with newlines taken into account
-        summary_len = sum(len(part) for part in summary) + len(summary) - 1
-
-        # Determine the space remaining after the summary. Reserve some extra
-        # space so we can at least report that all details were omitted if the
-        # message is too long.
-        if max_length is not None:
-            remaining_len = max_length - summary_len - 40
-
         if summary and show_footer:
             footer = ('--- ',
                       '%s %s, %s  ' % (urlwatch.pkgname, urlwatch.__version__, urlwatch.__copyright__),
                       'Website: %s  ' % (urlwatch.__url__,),
                       'watched %d URLs in %d seconds' % (len(self.job_states), self.duration.seconds))
-            footer_len = sum(len(part) for part in footer) + len(footer) - 1
+        else:
+            footer = None
 
-            if max_length is not None:
-                remaining_len -= footer_len
+        if not show_details:
+            details = None
 
-        # Calculate approximate available length per item, shared equally.
-        if max_length is not None:
-            len_per_details = remaining_len // len(details)
+        trimmed_msg = "*Parts of the report were omitted due to message length.*\n"
+        max_length -= len(trimmed_msg)
+
+        trimmed, summary, details, footer = MarkdownReporter._render(
+            max_length, summary, details, footer
+        )
 
         if summary:
-            yield from ('%d. %s' % (idx + 1, line) for idx, line in enumerate(summary))
+            yield from summary
             yield ''
 
         if show_details:
-            if max_length is not None and len_per_details < 100:
-                yield "Details omitted due to message length."
-            else:
-                unprocessed = len(details)
+            for header, body in details:
+                yield header
+                yield body
+                yield ''
 
-                for header, body in details:
-                    # Calculate the available length for the body and render it
-                    if max_length is not None:
-                        avail_length = len_per_details - len(header) - 1
-                    else:
-                        avail_length = None
-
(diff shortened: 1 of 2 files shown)


lib/urlwatch/reporters.py
```diff
@@ -668,18 +668,75 @@ def submit(self):
...
+
+        # Message to print when the diff is too long.
+        trim_message = "[... diff trimmed ...]"
+        trim_message_length = len(trim_message)
+
+        if max_length is None or len(s) + wrapper_length <= max_length:
+            return "```diff\n{}\n```".format(s)
+        else:
+            while len(s) + wrapper_length + trim_message_length > max_length:
+                s = re.sub("\n.*$", "", s)
```
thp, line 737: What if there's no newline in the string anymore (e.g. because there's a really really long line somewhere), would this be an infinite loop?

lib/urlwatch/reporters.py
```diff
@@ -668,18 +668,75 @@ def submit(self):
...
+        if summary and show_footer:
+            footer = ('--- ',
+                      '%s %s, %s  ' % (urlwatch.pkgname, urlwatch.__version__, urlwatch.__copyright__),
+                      'Website: %s  ' % (urlwatch.__url__,),
+                      'watched %d URLs in %d seconds' % (len(self.job_states), self.duration.seconds))
+            footer_len = sum(len(part) for part in footer) + len(footer) - 1
+            remaining_len -= footer_len
+
+        # Calculate approximate available length per item, shared equally.
+        len_per_details = remaining_len // len(details)
```
thp, line 688: Would/should shorter summaries give back to the pool? e.g. if there's one change that only has a short summary, then `len_per_details` can in general be longer?

thp (changes requested):
See comments. I do like the idea, but not sure if the approach chosen is the best.

lib/urlwatch/reporters.py
```diff
@@ -668,18 +668,75 @@ def submit(self):
...
-                        '%s %s, %s  ' % (urlwatch.pkgname, urlwatch.__version__, urlwatch.__copyright__),
-                        'Website: %s  ' % (urlwatch.__url__,),
-                        'watched %d URLs in %d seconds' % (len(self.job_states), self.duration.seconds))
+            yield from footer
+
+    @staticmethod
+    def _format_details_body(s, max_length):
+        # Number of characters for the Markdown code block wrapper (three
+        # backticks and a newline, twice).
+        wrapper_length = 8
```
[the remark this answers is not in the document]
dkasak: Yes. Added `diff` later on and forgot to take it into account. I was avoiding calculating the length dynamically each time, but I guess it's run rarely enough that it doesn't really matter.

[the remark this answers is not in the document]
dkasak: In any case, it's been corrected.

lib/urlwatch/reporters.py
dkasak: Oops, you're right. I modified this so that this problem cannot happen anymore.

lib/urlwatch/reporters.py
dkasak: At this point the length of the summaries should be accurately accounted for so they never take away more from the pool than they have to. The details do give back to the pool [here](https://github.com/thp/urlwatch/pull/572/files/38c15ac5004071f028e281c3ce657297fed863eb#diff-4a9994ac9199b1bd6143a8766ec1eb48R711), if they don't use the entirety of their allocated length.

Unless I misunderstood something.

dkasak:
I think I addressed your comments now and I also fixed the case when the limit is removed. Do you have any other suggestions?

lib/urlwatch/reporters.py
```diff
@@ -668,18 +668,88 @@ def submit(self):
...
+
+        # Message to print when the diff is too long.
+        trim_message = "[... diff trimmed ...]"
+        trim_message_length = len(trim_message)
+
+        if max_length is None or len(s) + wrapper_length <= max_length:
+            return "```diff\n{}\n```".format(s)
+        else:
+            while len(s) + wrapper_length + trim_message_length > max_length and '\n' in s:
+                s = re.sub("\n.*$", "", s)
```
thp, line 745: Couldn't we also use `.rfind('\n')` on the string to find the cutoff point without having to loop and remove line-by-line? Given that we know what the maximum length is, we can probably even do it with a single call (basically "find the last `\n` that comes before `max_length - trim_message_length - wrapper_length`, right?)

Here's some untested code (might contain off-by-one errors...):

```python3

target_max_length = max_length - trim_message_length - wrapper_length
pos = s.rfind('\n', 0, target_max_length)
if pos == -1:
    # we have a single long line so just cut it short
    pos = target_max_length
s = s[:pos]

File: lib/urlwatch/reporters.py [hunk 5]
```diff
@@ -668,18 +668,88 @@ def submit(self):
...
             details.extend(details_part)
 
+        # Sum of the lengths of parts, with newlines taken into account
+        summary_len = sum(len(part) for part in summary) + len(summary) - 1
+
+        # Determine the space remaining after the summary. Reserve some extra
+        # space so we can at least report that all details were omitted if the
+        # message is too long.
+        if max_length is not None:
+            remaining_len = max_length - summary_len - 40
```
thp, line 678: Could you combine all the `max_length is not None` blocks into a single block below?

Or maybe alternatively move the max length code into its own function that takes all the details part and does the length distribution?

thp: In other words, can you make it so that the code is easy to read/understand if `max_length is None` (right now one has to read lots of code and all the `max_length` conditions are not hit).

thp (changes requested):
Getting there! Some more comments...

lib/urlwatch/reporters.py
dkasak: That's better. I didn't think of it because I forgot about rfind's third argument. I'll handle it when I get some free time.

lib/urlwatch/reporters.py
dkasak: This block is here on its own because we're using `remaining_len` in the block below when we calculate the footer length so we need it set by then.

I don't like the current state of the code either, I'll see what I can do. Some tests might be nice too since this is tricky.

lib/urlwatch/reporters.py
dkasak: Done.

lib/urlwatch/reporters.py
dkasak: I refactored the code into a separate `_render` method which renders the report parts, trimming them down if necessary and reporting whether anything was trimmed. The code is now much easier to follow (IMO).

dkasak:
Sorry this took a while, I've been busy with other stuff.

lib/urlwatch/reporters.py
```diff
@@ -668,18 +668,144 @@ def submit(self):
...
 
         if summary and show_footer:
-            yield from ('--- ',
-                        '%s %s, %s  ' % (urlwatch.pkgname, urlwatch.__version__, urlwatch.__copyright__),
-                        'Website: %s  ' % (urlwatch.__url__,),
-                        'watched %d URLs in %d seconds' % (len(self.job_states), self.duration.seconds))
+            yield from footer
+
+    @staticmethod
+    def _render(max_length, summary=None, details=None, footer=None):
```
thp, line 706: Make this a `@classmethod` (see below)

lib/urlwatch/reporters.py
```diff
@@ -668,18 +668,144 @@ def submit(self):
...
+                    body_len_per_details = remaining_len // len(details)
+
+                    trimmed_details = []
+                    unprocessed = len(details)
+
+                    for header, body in details:
+                        # Calculate the available length for the body and render it
+                        avail_length = body_len_per_details - 1
+
+                        body_trimmed, body = MarkdownReporter._format_details_body(body, avail_length)
```
thp, line 763: Since you are referencing `MarkdownReporter` here, this should be a `@classmethod` and you should then use `cls._format_details_body()` here.

thp (changes requested):
Some finishing touches:

- [ ] Mention the changes in `CHANGELOG.md`
- [ ] See comments (classmethod / calling class member function)

thp (approved):
Thanks!

PR #590, opened 2020-11-20
feat(UrlJob): add url_for_human option

huxiba:
when url is api, use url_for_human in report, so click the link in report will show the content that we want.


thp (changes requested):
Good idea, but I'm not too happy with the name. Maybe `url_in_reports`? Or `report_url` or `url_for_report`?

thp:
Implemented slightly differently (`user_visible_url`) but basically the same idea.

PR #593, opened 2020-11-24
Update version check for python

v-m-r:
https://github.com/thp/urlwatch/blob/dfc5751bf3e42e0718b60c9aa0f5188f6c93c6bf/setup.py#L25
Updated Python dependency was declared as 3.5, but the version check still warns only for version < 3.3
This commit fixes it


thp (approved):
Nice find, thanks!

PR #600, opened 2020-12-05
fix filter bug

monperrus:
Fix crash:

```
  File "/home/martin/.local/lib/python3.5/site-packages/urlwatch/handler.py", line 88, in process
    data = FilterBase.auto_process(self, data)
  File "/home/martin/.local/lib/python3.5/site-packages/urlwatch/filters.py", line 84, in auto_process
    data = filter_instance.filter(data)
TypeError: filter() missing 1 required positional argument: 'subfilter'

```

### DIFF
--- a/lib/urlwatch/filters.py
+++ b/lib/urlwatch/filters.py
@@ -82,7 +82,8 @@ class FilterBase(object, metaclass=TrackSubClasses):
             filter_instance = filtercls(state.job, state)
             if filter_instance.match():
                 logger.info('Auto-applying filter %r to %s', filter_instance, state.job.get_location())
-                data = filter_instance.filter(data)
+                # filters require a subfilter argument
+                data = filter_instance.filter(data, None)
 
         return data
 


thp (approved):
This might interfere with existing user-defined filters, but yeah we should make it so that all filters take a subfilter.

thp:
Related documentation changes: https://github.com/thp/urlwatch/commit/fe4e766bd9e366068cded4f850ff508f3e0f8b3c

monperrus:
Thanks Thomas.

sbraz:
Hi, it would have been nice to emphasise the fact that this is a breaking change in the changelog. I quickly guessed that `TypeError: filter() takes 2 positional arguments but 3 were given` was caused by a missing `subfilter` parameter but I didn't expect breaking changes within a minor release.

thp:
Documented here: https://urlwatch.readthedocs.io/en/latest/deprecated.html#filters-without-subfilters-since-2-22

sbraz:
Hi Thomas, that's good to know. I wasn't aware of that page.

I still think breaking changes should have their own subsection in the changelog. Something that separates them from normal changes, maybe with a little :warning: icon.

thp:
> I still think breaking changes should have their own subsection in the changelog. Something that separates them from normal changes, maybe with a little ⚠️ icon.

Feel free to update the changelog to reflect this and make a PR. It's fine to improve the changelog also for already-released versions, even if this means it's only visible in Git and future versions and not with the changelog file shipped with the release.

PR #599, opened 2020-12-01
Basic colorization for wdiff-style differences in HtmlReporter (previ…

kongomongo:
…ously only in StdoutReporter)

Hi there. Small pull request to get the wdiff style colorization in HtmlReporter (Emails) in addition to the terminal colorized output.

PR #615, opened 2021-01-17
WDIFF multiline highlighting

kongomongo:
hi there, some time ago I made a PR (https://github.com/thp/urlwatch/pull/599) for basic WDIFF highlighting in Email reporter. However, since _diff_to_html() currently works line based with a generator, a multiline WDIFF like this won't be highlighted:

```
 Range (est.)        168.29 km
   State of Charge     [-78% (79%)
   Outside Temperature 2.5 °C
   Inside Temperature  3.2 °C-] {+62% (63%)+}
```

Since I am not very python-savy I could not integrate the fix while keeping the splitlines/yield syntax, which is probably good for performance. If someone could point out if my changes would be OK or terrible I'd be very glad. Many thanks in advance.

PS: Before, every line was enclosed with a span of class unified_nor, then every highlights were nested within with another span tag. With this fix, this will be changed by a single(!) unified_nor span around everything, and then lines beginning with +/- will be nested spans as well as WDIFF (even multiline) ones

### DIFF
--- a/lib/urlwatch/reporters.py
+++ b/lib/urlwatch/reporters.py
@@ -202,17 +202,14 @@ class HtmlReporter(ReporterBase):
                     url=urlwatch.__url__, count=len(self.job_states), duration=self.duration.seconds)
 
     def _diff_to_html(self, unified_diff):
-        for line in unified_diff.splitlines():
-            if line.startswith('+'):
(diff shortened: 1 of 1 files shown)


thp (approved):

thp:
Merged, thanks!

PR #619, opened 2021-02-05
Add options that require an argument as separate list elements

### LINKED ISSUE #618: Incorrect building of command arguments
https://github.com/thp/urlwatch/blob/4eb9391829267f42e669988a91cff6cbe112d067/lib/urlwatch/html2txt.py#L101

When building extra arguments for the command to be invoked for the html2text filter, options that take an argument (for example, -width 132), are put into a single elements of the "cmd" list. This results in a subprocess.Popen() invocation that takes this cmd list:

2021-02-05 12:33:37,880 html2txt DEBUG: Command: ['html2text', '-nobs', '-utf8', '-width 132'], stdout encoding: utf-8

instead of this:

2021-02-05 12:33:37,880 html2txt DEBUG: Command: ['html2text', '-nobs', '-utf8', '-width', '132'], stdout encoding: utf-8

The html2text command then fails because '-width 132' is not a valid option; only '-width' is.

peloy:
When building a command argument list for subprocess.Popen(), add
options that require an argument as separate list elements. This results
in a list like this:

2021-02-05 12:33:37,880 html2txt DEBUG: Command: ['html2text', '-nobs', '-utf8', '-width', '132'], stdout encoding: utf-8

instead of:

2021-02-05 12:33:37,880 html2txt DEBUG: Command: ['html2text', '-nobs', '-utf8', '-width 132'], stdout encoding: utf-8

which produces a CLI options parsing error because there is no option
called '-width 132'.

Resolves #618

### DIFF
--- a/lib/urlwatch/html2txt.py
+++ b/lib/urlwatch/html2txt.py
@@ -98,7 +98,7 @@ def html2text(data, baseurl, method, options):
 
     for k, v in options.items():
         if v is not None:
-            cmd.append('-{} {}'.format(k, v))
+            cmd.extend(['-{}'.format(k), '{}'.format(v)])
         else:
             cmd.append('-{}'.format(k))
 


thp:
Merged, thanks!

PR #620, opened 2021-02-07
Migrate to matrix-nio

### LINKED ISSUE #542: Matrix client dependency
When I look at the github page for the dependency of matrix_client at https://github.com/matrix-org/matrix-python-sdk it reads:

> We strongly recommend using the matrix-nio library rather than this sdk. It is both more featureful and more actively maintained.

Is there a plan to change to the matrix-nio library as a client dependency?

doubleloop:
Fixes #542

### DIFF
--- a/docs/source/deprecated.rst
+++ b/docs/source/deprecated.rst
@@ -29,23 +29,7 @@ Here is a sample configuration after update:
       user: @botusername:matrix.org
       enabled: true
 
-Then you need to manually trigger connection to obtain credentials and
-generate encryption keys:
-
-.. code:: bash
-
-    $ urlwatch --test-reporter matrix
-
-This is one time process during which you will be prompted for a password
-to the bot account (password will not be stored).
-Obtained credentials and encryption keys will be stored in user data directory
-approprite for your system (e.g. ``~/.local/share/urlwatch/`` on linux).
-
-Note that it is now not recommended to store ``access_token`` in configuration
-file and it will be only used if ``urlwatch`` recognizes configuration
-compatible with ``matrix_client`` (missing ``user`` field and containing
-``access_token`` field).
-
+For more details, check :ref:`Matrix reporter documentation<matrix_reporter>`
 
 Filters without subfilters (since 2.22)
 ---------------------------------------
--- a/docs/source/deprecated.rst
+++ b/docs/source/deprecated.rst
@@ -5,6 +5,48 @@ This page lists the features that are deprecated and steps to
 update your configuration to use the replacements (if any).
 
 
+matrix_client optional dependency (since 2.24)
+----------------------------------------------
+In older versions ``matrix_client`` was used as matrix provider.
+This optional dependency is now replaced with ``matrix-nio`` which is both
+more featureful (support encrypted rooms) and more actively maintained.
+
+Configuration is backwards compatible so after updating to 2.23 you
+only need to install ``matrix-nio`` and everything should work as previously.
+
+To use encryption you need to update configuration in following ways:
+
+#. Remove ``access_token`` field
+#. Add ``user`` field
+
+Here is a sample configuration after update:
+
+.. code:: yaml
+
+    matrix:
+      homeserver: https://matrix.org
+      room_id: "!roomroomroom:matrix.org"
+      user: @botusername:matrix.org
+      enabled: true
+
+Then you need to manually trigger connection to obtain credentials and
+generate encryption keys:
+
+.. code:: bash
+
+    $ urlwatch --test-reporter matrix
+
+This is one time process during which you will be prompted for a password
+to the bot account (password will not be stored).
+Obtained credentials and encryption keys will be stored in user data directory
+approprite for your system (e.g. ``~/.local/share/urlwatch/`` on linux).
+
+Note that it is now not recommended to store ``access_token`` in configuration
+file and it will be only used if ``urlwatch`` recognizes configuration
+compatible with ``matrix_client`` (missing ``user`` field and containing
+``access_token`` field).
+
+
 Filters without subfilters (since 2.22)
 ---------------------------------------
 
--- a/docs/source/deprecated.rst
+++ b/docs/source/deprecated.rst
@@ -9,7 +9,7 @@ matrix_client optional dependency (since 2.24)
 ----------------------------------------------
 In older versions ``matrix_client`` was used as matrix provider.
 This optional dependency is now replaced with ``matrix-nio`` which is both
-more featureful (support encrypted rooms) and more actively maintained.
+more featureful (supports encrypted rooms) and more actively maintained.
 
 Configuration is backwards compatible so after updating to 2.23 you
 only need to install ``matrix-nio`` and everything should work as previously.
--- a/docs/source/reporters.rst
+++ b/docs/source/reporters.rst
@@ -250,7 +250,7 @@ Here is a sample configuration:
      user: @botusername:matrix.org
      enabled: true
 
-Then you need to manually trigger connection to the matrix from urlparse
+Then you need to manually trigger connection to the ``Matrix`` from urlparse
 to obtain ``access_token`` and generate encryption keys:
 
 .. code:: bash
--- a/docs/source/reporters.rst
+++ b/docs/source/reporters.rst
@@ -231,21 +231,13 @@ You can have notifications sent to you through the `Matrix protocol`_.
 To achieve this, you first need to register a Matrix account for the bot
 on any homeserver.
 
-You then need to acquire an access token and room ID, using the
-following instructions adapted from `this
-guide <https://t2bot.io/docs/access_tokens/>`__:
+You then need to acquire room ID, using the following instructions:
 
-1. Open `Riot.im <https://riot.im/app/>`__ in a private browsing window
-2. Register/Log in as your bot, using its user ID and password.
-3. Set the display name and avatar, if desired.
-4. In the settings page, select the "Help & About" tab, scroll down to the bottom and click Access
-   Token: <click to reveal>.
-5. Copy the highlighted text to your configuration.
-6. Join the room that you wish to send notifications to.
-7. Go to the Room Settings (gear icon) and copy the *Internal Room ID*
-   from the bottom.
-8. Close the private browsing window **but do not log out, as this
-   invalidates the Access Token**.
+#. Open `Riot.im <https://riot.im/app/>`__ in a private browsing window
+#. Register/Log in as your bot, using its user ID and password.
+#. Set the display name and avatar, if desired.
+#. Join the room that you wish to send notifications to.
+#. Go to the Room Settings (gear icon) and copy the *Internal Room ID* from the bottom.
 
 Here is a sample configuration:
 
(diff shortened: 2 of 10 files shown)


thp (changes requested):
This probably needs a section in the "Deprecated features" documentation on how to migrate from an old matrix configuration to the new one, otherwise people will have their matrix notifications broken (either silently or loudly), which isn't nice. Is there a way we can auto-migrate from the old way to the new way, so Matrix notifications just keep running? Alternatively, should this be a "matrix-nio" reporter and we leave support for the old one in place, but the documentation will just mention that a matrix reporter existed, but new installations should use "matrix-nio"?

thp:
Any news on this? Do you think you want to work on getting this ready to be merged?

doubleloop:
>needs a section in the "Deprecated features" documentation on how to migrate from an old matrix configuration to the new one, otherwise people will have their matrix notifications broken (either silently or loudly), which isn't nice.

I added special case so that if configuration compatible with `matrix_client` is found, it will still work as expected (with no encryption support). I also added `Deprecated Features` section describing changes required for encryption support.

>Do you think you want to work on getting this ready to be merged?
Yes, waiting for some review/feedback and list of things needed to make this ready to be merged.

docs/source/deprecated.rst
```diff
@@ -5,6 +5,48 @@ This page lists the features that are deprecated and steps to
 update your configuration to use the replacements (if any).
 
 
+matrix_client optional dependency (since 2.24)
+----------------------------------------------
+In older versions ``matrix_client`` was used as matrix provider.
+This optional dependency is now replaced with ``matrix-nio`` which is both
+more featureful (support encrypted rooms) and more actively maintained.
```
thp, line 12: "support encrypted rooms" -> "supports encrypted rooms"

docs/source/deprecated.rst
```diff
@@ -5,6 +5,48 @@ This page lists the features that are deprecated and steps to
...
+
+This is one time process during which you will be prompted for a password
+to the bot account (password will not be stored).
+Obtained credentials and encryption keys will be stored in user data directory
+approprite for your system (e.g. ``~/.local/share/urlwatch/`` on linux).
+
+Note that it is now not recommended to store ``access_token`` in configuration
+file and it will be only used if ``urlwatch`` recognizes configuration
+compatible with ``matrix_client`` (missing ``user`` field and containing
+``access_token`` field).
```
thp, line 47: Just replace this with a link to the Matrix reporter documentation (merging anything that's here and not there to there).

docs/source/reporters.rst
```diff
@@ -231,33 +231,37 @@ You can have notifications sent to you through the `Matrix protocol`_.
...
-3. Set the display name and avatar, if desired.
-4. In the settings page, select the "Help & About" tab, scroll down to the bottom and click Access
-   Token: <click to reveal>.
-5. Copy the highlighted text to your configuration.
-6. Join the room that you wish to send notifications to.
-7. Go to the Room Settings (gear icon) and copy the *Internal Room ID*
-   from the bottom.
-8. Close the private browsing window **but do not log out, as this
-   invalidates the Access Token**.
+You then need to acquire room ID, using the following instructions:
```
thp, line 234: "to acquire room ID" -> "to acquire the room ID"

docs/source/reporters.rst
```diff
@@ -231,33 +231,37 @@ You can have notifications sent to you through the `Matrix protocol`_.
...
+to obtain ``access_token`` and generate encryption keys:
+
+.. code:: bash
+
+    $ urlwatch --test-reporter matrix
+
+This is one time process during which you will be prompted for a password
+to the bot account (password will not be stored).
+Obtained credentials and encryption keys will be stored in user data directory
+approprite for your system (e.g. ``~/.local/share/urlwatch/`` on linux).
```
thp, line 262: "approprite" -> "appropriate"

thp: Remove "(e.g. ``~/.local/share/urlwatch/`` on linux)" and replace with something like "(relative to urlwatch's configuration directory)".

lib/urlwatch/tests/test_handler.py
```diff
@@ -90,7 +90,7 @@ def test_pep8_conformance():
 class ConfigForTest(BaseConfig):
     def __init__(self, config, urls, cache, hooks, verbose):
         (prefix, bindir) = os.path.split(os.path.dirname(os.path.abspath(sys.argv[0])))
-        super().__init__('urlwatch', os.path.dirname(__file__), config, urls, cache, hooks, verbose)
+        super().__init__('urlwatch', os.path.dirname(__file__), os.path.dirname(__file__), config, urls, cache, hooks, verbose)
```
thp, line 93: Should this use a temporary folder instead of the path where test_handler is? If it's installed, it might not be writable to a user wanting to run the tests.

docs/source/deprecated.rst
```diff
@@ -5,6 +5,48 @@ This page lists the features that are deprecated and steps to
...
+In older versions ``matrix_client`` was used as matrix provider.
+This optional dependency is now replaced with ``matrix-nio`` which is both
+more featureful (support encrypted rooms) and more actively maintained.
+
+Configuration is backwards compatible so after updating to 2.23 you
+only need to install ``matrix-nio`` and everything should work as previously.
+
+To use encryption you need to update configuration in following ways:
+
+#. Remove ``access_token`` field
```
thp, line 19: Should we warn when loading a config that still has `access_token` set?

lib/urlwatch/reporters.py
```diff
@@ -935,44 +938,129 @@ def _format_output(self, job_state):
...
+                extras=["fenced-code-blocks", "highlightjs-lang"]).convert(
+                    body_markdown)
+
+            content["format"] = "org.matrix.custom.html"
+            content["formatted_body"] = body_html
         else:
-            logger.debug('Not formatting as Markdown; dependency on markdown2 not met?')
-            client_api.send_message(room_id, body_markdown)
+            logger.debug(
+                'Not formatting as Markdown; dependency on markdown2 not met?')
```
thp, line 1029: I know this log message has been like this before, but since we touch it, maybe we can just make it something like:

    markdown2 could not be imported, skipping Markdown formatting

lib/urlwatch/handler.py
```diff
@@ -174,8 +174,9 @@ def _generate_diff(self):
 
 
 class Report(object):
-    def __init__(self, urlwatch_config):
-        self.config = urlwatch_config.config_storage.config
+    def __init__(self, urlwatch_obj):
+        self.config = urlwatch_obj.config_storage.config
+        self.urlwatch_config = urlwatch_obj.urlwatch_config
```
thp, line 179: Is this change necessary?

thp: Or in other words, could we just set the data directory as additional parameter, not expose the whole config object?

thp (changes requested):
See comments. We're getting there :)

docs/source/deprecated.rst
doubleloop: https://github.com/thp/urlwatch/pull/620/files#diff-5ee1abebe5480db745909e862aaa226d8f66babc4ab506b526c49cf7c0059fe0R1034-R1038

docs/source/deprecated.rst
doubleloop: could you verify if I have done it correctly?

lib/urlwatch/tests/test_handler.py
doubleloop: Made temporary folder. Not sure why `os.path.dirname(__file__)` is used here.
note: what do you thing about renaming `urlwatch_dir` to `urlwatch_config_dir` ?

lib/urlwatch/handler.py
doubleloop: Not sure why not to expose the whole config (other features might require access to command line args or other paths like cache), but I will change this.

> could we just set the data directory as additional parameter
Note that I only renamed existing parameter and did not introduce any new one (`urlwatch_config` -> `urlwatch_obj`). Previous name was misleading (https://github.com/thp/urlwatch/blob/master/lib/urlwatch/main.py#L54)

doubleloop:
I rebased on master and added `review fix` commits (I plan to squash them before merge).

lib/urlwatch/reporters.py
```diff
@@ -935,44 +938,129 @@ def _format_output(self, job_state):
...
+                            'Could not read cred json file, overwriting.')
+                self.load_store()
+
+            if not self.user_id or not self.device_id or not self.access_token:
+                logger.debug("Generating new credentials")
+                print("### Generating new Matrix credentials. ###")
+                device_name = input(f"Choose a name for this device: [{self.DEFAULT_DEVICE_NAME}] ")
+                if not device_name:
+                    device_name = self.DEFAULT_DEVICE_NAME
+                pw = getpass.getpass(f"Matrix password for {self.user} (it will not be stored): ")
```
thp, line 985: This should most likely be a separate mode from the command line, as we don't want random invocations to be interactive (urlwatch is most likely running as a cronjob or via some other non-interactive means).

thp: See the `--xmpp-login` command-line option for an example. Maybe combine it like `--test-filter` so that you can do e.g. `--login-reporter xmpp` and `--login-reporter matrix` (as `--xmpp-login` is an interactive thing, maybe just update docs + mention it in the list of deprecated things that `--xmpp-login is now `--login-reporter xmpp`).

lib/urlwatch/reporters.py
```diff
@@ -935,44 +938,129 @@ def _format_output(self, job_state):
...
+
+            client = UrlwatchMatrixClient(self.config, self.report.urlwatch_data_dir)
+            try:
+                await client.login()
+                if room_id not in client.rooms:
+                    await client.join(room_id)
+                    logger.debug(f'joined room {room_id}')
+                if client.should_upload_keys:
+                    await client.keys_upload()
+                await client.sync(timeout=30000, full_state=True)
```
thp, line 1055: `30000` should probably be an named constant (`MATRIX_CLIENT_SYNC_TIMEOUT_MS`?)

docs/source/reporters.rst
```diff
@@ -231,33 +232,37 @@ You can have notifications sent to you through the `Matrix protocol`_.
...
-   Token: <click to reveal>.
-5. Copy the highlighted text to your configuration.
-6. Join the room that you wish to send notifications to.
-7. Go to the Room Settings (gear icon) and copy the *Internal Room ID*
-   from the bottom.
-8. Close the private browsing window **but do not log out, as this
-   invalidates the Access Token**.
+You then need to acquire the room ID, using the following instructions:
+
+#. Open `Riot.im <https://riot.im/app/>`__ in a private browsing window
```
thp, line 237: Is there some command-line matrix client that can be used for this without relying on `riot.im` or a similar service? I know that this was in here before, but since one of the advantages of this change is that rooms are now encrypted, the encryption would be kinda pointless if initially we rely on trusting `riot.im`.

thp (changes requested):
See comments. We're getting there!

docs/source/reporters.rst
doubleloop: > Is there some command-line matrix client that can be used for this without relying on riot.im or a similar service?

According to spec, room `name` is not unique, room `alias` is optional so `room ID` in general is the only reliable room identifier (https://matrix.org/docs/spec/client_server/latest#m-room-name). `room ID` is obtained during room creation and stored by the client. Any client used to create room should be able to display `room ID` (I can note in docs that ~riot.im~ element is just an example)

>the encryption would be kinda pointless if initially we rely on trusting riot.im

Why?

thp:
@doubleloop Any updates on this PR?

thp:
Closing for now. Please reopen once rebased.

Issue #592, opened 2020-11-24
Add filter to parse json and select json path


piejanssens:
https://www.journaldev.com/33265/python-jsonpath-examples#:~:text=JSONPath%20is%20an%20expression%20language,read%20the%20complete%20JSON%20data.

piejanssens:
Solved using `  - shellpipe: "python3 -c \"import sys, json; print(json.load(sys.stdin)['data'])\""
`

PR #626, opened 2021-03-03
Add jq filter

robgmills:
Adds the ability to select content from JSON data using `jq` - a lightweight JSON processor - filter syntax.  The `jq` filters are similar to CSS or XPath selectors, with added operators to transform the data.

Addresses #592 in a different way.


robgmills:
Thanks for sharing this project with us!  I'm not super well-versed in Python projects and this has been a good experience.

I can see in the Travis job how to run tests and that I have some syntax issues in the documentation.  I'll add a test or tests for this filter - is there anything in particular you want me to test for?  I'll also clean up the documentation syntax issues I introduced.

robgmills:
These documentation driven tests are pretty cool!  I'll have to use them as inspiration later.

Sorry for the disjointed commits - I normally try to keep things cleaner.  This was a good learning experience!  I'm excited to hear your thoughts!

lib/urlwatch/filters.py
```diff
@@ -855,3 +860,29 @@ def filter(self, data, subfilter):
...
+        except ValueError:
+            raise ValueError('The url response contained invalid JSON')
+
+        if 'query' not in subfilter:
+            raise ValueError('{} filter needs a query'.format(self.__kind__))
+
+        if jq is None:
+            raise ImportError('Please install jq')
+
+        return jq.text(subfilter['query'], json.loads(data))
```
thp, line 888: One "optimization" here could be to store the result of `json.loads(data)` in line 878, and re-use the result in line 888 to not have to re-parse the JSON data.

thp (changes requested):
Nice and clean. One comment about avoiding double-parsing of JSON. Also, you could add this change to `CHANGELOG.md`.

robgmills:
Thanks for the feedback @thp!

thp (approved):

thp:
Merged, thanks!

smithl:
Appreciate this addition very much, nice work!

PR #627, opened 2021-03-04
add gitlab tag filter to docs

julianuu:


docs/source/filters.rst
```diff
@@ -385,6 +385,15 @@ project for the latest release version, to be notified of new releases:
      - html2text: re
      - strip
 
+This is the corresponding version for Gitlab tags:
+
+.. code:: yaml
+
+   url: https://gitlab.com/chinstrap/gammastep/-/tags
```
thp, line 392: Since we run automated tests on the filter documentation, you need to add the (simplified) content of this URL and the expected output to `lib/urlwatch/tests/data/filter_documentation_testdata.yaml`, just provide some simplified HTML snippet in the "input" key, and the expected output in the "output" key.

You can run the tests with "pytest -v". You can either install some dependencies that it needs (tesseract, pdftotext, ...) or just ignore those failures and check if the `lib/urlwatch/tests/test_filter_documentation.py` tests work properly.

docs/source/filters.rst
```diff
@@ -371,7 +371,7 @@ are separated by an empty line):
          separator: "\n\n"
 
 
-Watching Github releases
+Watching Github releases and Gitlab tags
```
thp, line 374: Maybe adjust the `------` line below to be of same length.

thp (changes requested):
See comments. Needs to have a test case added, then it can be merged.

docs/source/filters.rst
julianuu: Only installing `pytest` somehow wasn't enough to make it run. I didn't try to figure out the dependencies, but instead repeatedly pushed edits until the checks on travis-ci passed. I hope that was ok.

thp (approved):

thp:
Merged, thanks!

PR #633, opened 2021-04-04
Add reporter: Prowl App

nitz:
[Prowl](https://prowlapp.com) is a web-based push notification platform not unlike Pushover or Pushbullet.

I added it as it's the one I personally use, and thought perhaps someone else out there might get use from it too.


CHANGELOG.md
```diff
@@ -6,6 +6,10 @@ The format mostly follows [Keep a Changelog](http://keepachangelog.com/en/1.0.0/
 
 ## [UNRELEASED]
 
+### Added
+
+- New reporter: '[prowl](https://prowlapp.com)' (by nitz).
```
thp, line 11: You could use backticks here instead of single quotes.

lib/urlwatch/reporters.py
```diff
@@ -978,3 +978,65 @@ def submit(self):
 
         for chunk in chunkstring(text, self.MAX_LENGTH, numbering=True):
             asyncio.run(xmpp.send(chunk))
+
+class ProwlReport(TextReporter):
```
thp, line 982: The PEP-8 checker complains:

```
/home/travis/build/thp/urlwatch/lib/urlwatch/reporters.py:982:1: E302 expected 2 blank lines, found 1
```

(just add another blank line above the `class ProwlReport...` line.

Maybe also name the class "ProwlReporter" to be in line with other class names.

lib/urlwatch/reporters.py
```diff
@@ -978,3 +978,65 @@ def submit(self):
...
+    __kind__ = 'prowl'
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+    def submit(self):
+        api_base = 'https://api.prowlapp.com/publicapi/'
+        api = {
+            'add': api_base + 'add'
+        }
```
thp, line 994: If there's just a single endpoint, feel free to to just simplify this to:

```
   api_add = 'https://api.prowlapp.com/publicapi/add'
```

lib/urlwatch/reporters.py
```diff
@@ -978,3 +978,65 @@ def submit(self):
...
+
+        # 'subject' used in the config file, but the API
+        # uses what might be called the subject as the 'event'
+        event = self.config['subject'].format(**subject_args)
+
+        # 'application' is prepended to the message in prowl,
+        # to show the source of the notification. this too,
+        # is user configurable, and may reference subject args
+        application = self.config['application'].format(**subject_args)
```
thp, line 1015: Use this instead, so that the user can leave out the "application" config key:

```
    application = self.config.get('application')
    if application is not None:
        application = application.format(**subject_args)
    else:
        application = '{0} v{1}'.format(urlwatch.pkgname, urlwatch.__version__)
```

lib/urlwatch/reporters.py
```diff
@@ -978,3 +978,65 @@ def submit(self):
...
+
+        # if the user didn't set application, just use the package name and version.
+        if not application:
+            application = ('{0} v{1}'.format(urlwatch.pkgname, urlwatch.__version__))[:256].encode('utf8')
+
+        # build the data to post
+        post_data = {
+            'event': event[:1024].encode('utf8'),
+            'description': text[:10000].encode('utf8'),
+            'application': application[:256].encode('utf8'),
```
thp, line 1025: I'm pretty sure that you don't want to encode the data as UTF-8 here, as the requests posting code should take care of that, right?

thp (changes requested):
See comments.

nitz:
Fantastic feedback! (also I'm a bit new to Python, is there an easy way to run a PEP-8 (and other PEP) checkers locally?)

Okay! All those changes made, except the UTF-8 one, as I just wanted to confirm with you before I did: What I read in the requests [docs](https://docs.python-requests.org/en/latest/api/#behavioural-changes), it says it converts all the keys to UTF-8, and that values should avoid passing unicode, which makes it seem possible? The prowl api [does](https://www.prowlapp.com/api.php#methods) does specify UTF-8. The requests docs also mentions that _all_ values should be strings, so should I be stringifying `api_key` and `priority` too?

What do you think? Or I guess rather: what would you prefer :)

Thanks for the feedback and super expedant review!

thp (approved):

thp:
Merged, thanks!

Issue #331, opened 2018-12-06
Delay between requests


cjohnsonuk:
Where a URL list contains many pages on the same site would it be possible to add an arbitrary delay (possibly random) between consecutive requests?  Our own web page has a software firewall that prevents access from addresses making more than, for example, 20 requests per minute.  If my watch process takes 10 minutes rather than 10 seconds that is not an issue, as long as it checks all the pages and doesn't get blocked.  Something like a 
`PauseBetweenRequests : 10 seconds`
in the config file would allow checking of sites that have countermeasures for scripted requests.

cfbao:
This seems difficult to implement within the current structure, and possibly a bit out of scope.

I believe it's in urlwatch's design to finish its jobs as quickly as possible, and leave long term scheduling issues to the scheduler and other tools.

nille02:
How about a delay in the jobs. For example to set a interval for each job with each urlwatch run we exclude all jobs that didn't med the timeframe. 

interval: 86400 # Check only after 24 Hours
interval:  86400-172800 # Checks after 24 Hours up to 48 Hours

So if you run Multiple Times, Jobs that are already checked would be excluded. 

It would not fix the Problem in general but you can increase the urlwatch runs and decrease amount of requests

cfbao:
> How about a delay in the jobs. For example to set a interval for each job with each urlwatch run we exclude all jobs that didn't med the timeframe.

This is the same idea as #148, #171, and can be easily implemented once the cache database redesign #360 is merged.

However, this issue is different. To resolve this, we need to space out different jobs in the same run.

nille02:
But it would lessen the requests because you can dispense it over some time. I have a similar Website that blocks request after X requests per second. My Workaround currently is to split the url file with around 40 to 80 Jobs per File. And each Block is called with some time between them.

cfbao:
If you've already split your jobs into multiple urls files, then there's not really a need to set an interval in urlwatch. You can handle interval from the scheduler. Just like how your workaround works.

I believe this issue is about spacing out requests even if you've put everything into one urls file.

nille02:
How about a additional option. i would make a copy from list_urls in command.py and instead of print just run one job and use sleep() after each job. Its a small hack but it would be working.

nille02:
@cjohnsonuk you can try a automatchfilter from the hooks (CustomRegexMatchUrlFilter) and add a sleep in the filter. with this there are Max 10 requests at the same time. 

```python
import re
import random
[...]

Issue #171, opened 2017-09-17
Allow specifying individual check intervals for each item


n-st:
This issue is basically a clarification of #148:  
Like @strauhmanis, I want to check different sites/commands at different intervals.

So far I've resorted to having `urlwatch --urls=urlwatch.5mins.yml` and `urlwatch --urls=urlwatch.hourly.yml` in my crontab, but this approach doesn't scale well (I now want to add daily checks for some sites).

We already have a database as the back-end (though I don't know the specific layout), so couldn't we just keep a "run next check at or after *xx:yy*" field in there and run `urlwatch` from cron every five minutes?

PR #641, opened 2021-05-06
add command line argument to specify which jobs will be run

dbro:
Hi Thomas- this is a simple change that might be useful in situations where some jobs should be run at different intervals. For example, if only jobs 2,4,and 6 should be run hourly and all the rest should run once per day, then this pair of crontab lines would do that:

05 0-2,4-23   * * * /home/pi/.local/bin/urlwatch 2 4 6
05 3   * * * /home/pi/.local/bin/urlwatch

For small sets of jobs, this could be an adequate solution for the use cases described in issues #148 #171 #331 .

### DIFF
--- a/lib/urlwatch/tests/test_handler.py
+++ b/lib/urlwatch/tests/test_handler.py
@@ -11,7 +11,7 @@ import tempfile
 import os
 
 from urlwatch import storage
-from urlwatch.config import BaseConfig
+from urlwatch.config import CommandConfig
 from urlwatch.storage import YamlConfigStorage, CacheMiniDBStorage
 from urlwatch.main import Urlwatch
 from urlwatch.util import import_module_from_source
@@ -87,12 +87,10 @@ def test_pep8_conformance():
     assert result.total_errors == 0, "Found #{0} code style errors".format(result.total_errors)
 
 
-class ConfigForTest(BaseConfig):
+class ConfigForTest(CommandConfig):
     def __init__(self, config, urls, cache, hooks, verbose):
         (prefix, bindir) = os.path.split(os.path.dirname(os.path.abspath(sys.argv[0])))
-        super().__init__('urlwatch', os.path.dirname(__file__), config, urls, cache, hooks, verbose)
-        self.edit = False
-        self.edit_hooks = False
+        super().__init__('urlwatch', os.path.dirname(__file__), bindir, prefix, config, urls, hooks, cache, verbose)
 
 
 @contextlib.contextmanager
--- a/lib/urlwatch/worker.py
+++ b/lib/urlwatch/worker.py
@@ -53,10 +53,10 @@ def run_parallel(func, items):
 def run_jobs(urlwatcher):
     cache_storage = urlwatcher.cache_storage
     jobs = [job.with_defaults(urlwatcher.config_storage.config)
-            for job in urlwatcher.jobs]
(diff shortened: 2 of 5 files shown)


mborsetti:
This has been added to the `webchanges` fork: https://webchanges.readthedocs.io/en/stable/changelog.html#version-3-6-0

thp:
The job index is kind of volatile. Wonder if we should have e.g. an optional `groups` key on each job (a list of strings) and you can then run urlwatch only for a specific group?

dbro:
I think the syntax for this should match how the JOB is described for the commands --test-filter and --test-diff-filter . I noticed that they can match by either index numbers or (full) url values. However those commands are probably only used in developing and debugging job settings, and so they would only need to select one job at a time. The use case for specifying groups of jobs is not familiar to me, because I have only about 10 different jobs - so I don't think I should try and guess how to design a UI for it.

lib/urlwatch/tests/test_handler.py
```diff
@@ -87,12 +87,10 @@ def test_pep8_conformance():
...
 
 
-class ConfigForTest(BaseConfig):
+class ConfigForTest(CommandConfig):
     def __init__(self, config, urls, cache, hooks, verbose):
         (prefix, bindir) = os.path.split(os.path.dirname(os.path.abspath(sys.argv[0])))
-        super().__init__('urlwatch', os.path.dirname(__file__), config, urls, cache, hooks, verbose)
-        self.edit = False
-        self.edit_hooks = False
+        super().__init__('urlwatch', os.path.dirname(__file__), bindir, prefix, config, urls, hooks, cache, verbose)
```
thp, line 93: This change is unrelated to the PR?

lib/urlwatch/worker.py
```diff
@@ -53,10 +53,10 @@ def run_parallel(func, items):
 def run_jobs(urlwatcher):
     cache_storage = urlwatcher.cache_storage
     jobs = [job.with_defaults(urlwatcher.config_storage.config)
-            for job in urlwatcher.jobs]
+            for (idx, job) in enumerate(urlwatcher.jobs) if ((idx + 1) in urlwatcher.urlwatch_config.joblist or (len(urlwatcher.urlwatch_config.joblist) == 0))]
```
thp, line 56: Might be good before filtering the list of jobs to filter out:

    if not all(1 <= idx <= len(urlwatcher.jobs) for idx in urlwatcher.urlwatch_config.joblist):
        raise ValueError(f'All job indices must be between 1 and {len(urlwatcher.jobs)}: {urlwatcher.urlwatch_config.joblist}')

thp: Also `len(urlwatcher.urlwatch_config.joblist) == 0` should be the same as `not urlwatcher.urlwatch_config.joblist`

thp (changes requested):
See comments. I guess we could just put it in with the indexes for now. Will be hard to reshuffle this if the jobs file is edited, but that is up to the user then.

Can you add documentation for this feature too?

lib/urlwatch/tests/test_handler.py
dbro: This fixed some errors when running the tests, because with CommandConfig.__init__() the parse_args() function gets executed, but with the BaseConfig.__init__() it does not.

lib/urlwatch/worker.py
dbro: Good ideas. Should I create a new commit with these suggestions?

dbro:
Sure, done. The pull request has been updated with 2 new commits.

thp (approved):
Nice, thanks!

PR #646, opened 2021-05-30
telegram: Add silent messages, monospace formatting

ml-:


lib/urlwatch/storage.py
```diff
@@ -126,6 +126,8 @@
             'enabled': False,
             'bot_token': '',
             'chat_id': '',
+            'monospaced': False,
```
thp, line 129: Use `monospace` for this key name, not `monospaced`.

lib/urlwatch/storage.py
```diff
@@ -126,6 +126,8 @@
...
+            'disable_notification': False,
```
thp, line 130: Use `silent` instead of `disable_notification`.

lib/urlwatch/reporters.py
```diff
@@ -600,11 +600,29 @@ def submit(self):
...
+
     def submitToTelegram(self, bot_token, chat_id, text):
         logger.debug("Sending telegram request to chat id:'{0}'".format(chat_id))
-        result = requests.post(
-            "https://api.telegram.org/bot{0}/sendMessage".format(bot_token),
-            data={"chat_id": chat_id, "text": text, "disable_web_page_preview": "true"})
+
+        data = {"chat_id": chat_id,
+                "text": text,
+                "disable_notification": self.config['disable_notification'],
```
thp, line 613: Since this might not be set, you have to use:
```
self.config.get('disable_notification', False)
```

(or `'silent'`, see other comments)

lib/urlwatch/reporters.py
```diff
@@ -600,11 +600,29 @@ def submit(self):
...
+                "disable_web_page_preview": True}
+
+        if self.config['monospaced']:
```
thp, line 616: Since this might not be set, you have to use:
```
self.config.get('monospaced', False)
```

(or `monospace`, see other comments)

lib/urlwatch/reporters.py
```diff
@@ -600,11 +600,29 @@ def submit(self):
 
         return result
 
+    @staticmethod
+    def _format_body(text: str) -> str:
+        return "```diff\n{}\n```".format(
+            text.translate(str.maketrans({'`': r'\`', '\\': r'\\'})))
```
thp, line 606: Not sure if this is the best way to escape this (see [Markdown syntax](https://daringfireball.net/projects/markdown/syntax)). If you use ` ``` ` as the delimiter, you might just need to escape 3 literal backticks in one line, not sure...

thp (changes requested):
See comments.

lib/urlwatch/reporters.py
ml-: All `` ``` `` inside escapes just fine. First the translation is done, then wrapped `` ```diff\n{}\n``` `` . The [Telegram Bot API][1] only wants backticks and backslashes escaped inside when using MarkdownV2 parse_mode. I have found an example over at [python-telegram-bot/python-telegram-bot][2], but instead of running the input through the regexp library I used `str.maketrans` / `str. translate`.

Telegram Desktop
(excuse the large screenshots)

Screenshot_20210606_154655


[1]: https://core.telegram.org/bots/api#markdownv2-style
[2]: https://github.com/python-telegram-bot/python-telegram-bot/blob/master/telegram/utils/helpers.py#L149

lib/urlwatch/storage.py
ml-: renamed

lib/urlwatch/storage.py
ml-: renamed

lib/urlwatch/reporters.py
ml-: added

lib/urlwatch/reporters.py
ml-: added

thp (changes requested):
This looks good now. Please add entries to `CHANGELOG.md` and then this can be merged if CI is happy.

thp (approved):
Thanks!

Issue #555, opened 2020-08-18
Navigate jobs not working in docker container


muvaf:
When I run `urlwatch` in a docker container with a job whose `navigate` is given, I get the following error:
```
Exception while releasing resources for job: <browser navigate='<URL>' name='<JOBNAME>' filter=[{'xpath': '/html/body/div[4]/div[4]/form/div/div[3]/table/tbody'}, {'grepi': 'google-query'}, {'html2text': {'method': 'pyhtml2text', 'unicode_snob': True, 'body_width': 0, 'inline_links': False, 'ignore_links': True, 'ignore_images': True, 'pad_tables': False, 'single_line_break': True}}]>
Traceback (most recent call last):
  File "/usr/local/lib/python3.6/dist-packages/urlwatch/handler.py", line 77, in __exit__
    self.job.main_thread_exit()
  File "/usr/local/lib/python3.6/dist-packages/urlwatch/jobs.py", line 377, in main_thread_exit
    self.ctx.close()
AttributeError: 'BrowserJob' object has no attribute 'ctx'
===========================================================================
01. ERROR: <JOBNAME>
===========================================================================

---------------------------------------------------------------------------
ERROR: <JOBNAME AND URL>
---------------------------------------------------------------------------
Traceback (most recent call last):
  File "/usr/local/lib/python3.6/dist-packages/urlwatch/handler.py", line 67, in __enter__
    self.job.main_thread_enter()
  File "/usr/local/lib/python3.6/dist-packages/urlwatch/jobs.py", line 374, in main_thread_enter
    self.ctx = BrowserContext()
  File "/usr/local/lib/python3.6/dist-packages/urlwatch/browser.py", line 86, in __init__
    BrowserContext._BROWSER_LOOP = BrowserLoop()
  File "/usr/local/lib/python3.6/dist-packages/urlwatch/browser.py", line 45, in __init__
    self._browser = self._event_loop.run_until_complete(self._launch_browser())
  File "/usr/lib/python3.6/asyncio/base_events.py", line 484, in run_until_complete
    return future.result()
  File "/usr/local/lib/python3.6/dist-packages/urlwatch/browser.py", line 51, in _launch_browser
    browser = yield from pyppeteer.launch()
  File "/usr/local/lib/python3.6/dist-packages/pyppeteer/launcher.py", line 305, in launch
    return await Launcher(options, **kwargs).launch()
[...]

PR #652, opened 2021-07-09
Add option for browser jobs to disable Chromium sandbox

dbro:
I was getting errors when running a browser job which were fixed by adding the "--no-sandbox" option for Chromium. This pull requests adds support for this option in the yaml browser job definition.

This was mentioned in an early message as part of the discuss for issue #555 but I don't know if it is still relevant.

### DIFF
--- a/docs/source/advanced.rst
+++ b/docs/source/advanced.rst
@@ -272,6 +272,12 @@ before a page is considered loaded by using the `wait_until` option. It can take
   - `networkidle2` will wait until there are no more than 2 network connections for at least 500 ms.
 
 
+Disabling Chromium's sandbox
+----------------------------
+
+For browser jobs, you can disable the Chromium sandbox to isolate processes. BEWARE: This is a less secure method of loading web pages and should be used carefully. See https://chromium.googlesource.com/chromium/src/+/refs/heads/main/docs/design/sandbox.md for details.
+
+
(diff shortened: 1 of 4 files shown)


dbro:
Following up: this seems to have been the problem that was causing pyppeteer to fail:
https://github.com/puppeteer/puppeteer/blob/main/docs/troubleshooting.md#setting-up-chrome-linux-sandbox

so in my case instead of disabling the Chromium sandbox, I could run this command first:
sudo sysctl -w kernel.unprivileged_userns_clone=1

This is on Debian runing kernel version 4.19

mborsetti:
@dbro You may want to take a look at the [webchanges](https://pypi.org/project/webchanges/) fork, which was suggested by the author to optimize it for HTML; it has quite a few improvements specific to Pyppetter-run jobs (including, if you need it for your setup, the [optional directive](https://webchanges.readthedocs.io/en/stable/jobs.html#optional-directives) `switches` for Chromium that allows you to pass on any switch you want).  

See a list of improvements [here](https://webchanges.readthedocs.io/en/stable/migration.html) and examples using switches starting [here](https://webchanges.readthedocs.io/en/stable/advanced.html#using-a-chromium-revision-matching-a-google-chrome-chromium-release).

dbro:
Thanks. I will close this pull request.

PR #653, opened 2021-07-20
Add a uniq filter.

Sveder:


Sveder:
@thp wdyt?
I tried using `shellpipe: uniq` (didn't work) but makes sense to me to have this as a first class as `sort` exists.

thp:
The reason why `sort` exists is historical, there was no `shellpipe` filter when `sort` was introduced.

The reason why your `uniq` with shellpipe didn't work is most likely because you didn't sort it before (`uniq` in Unix requires lines to be sorted).

So for line-based uniqueness, I think it's fine to just sort it first (try `shellpipe: 'sort -u'` or `shellpipe: 'sort | uniq'`).

What might be interesting is a kind of "unique / sort" kind of filter that works on e.g. CSS or XPath selectors, as this is kind of hard to do with built-in unix commands.

Feel free to turn this PR into a documentation change that shows how to use shellpipe + `sort -u` or `sort | uniq` to filter duplicate lines.

Sveder:
@thp thanks for answering. You might be right that I didn't play with `shellpipe: uniq` enough, and indeed now that I read `uniq` docs it is not what I wanted as I don't want to sort the data. 
 
In this case, I think my implementation of `uniq` is definitely simpler and more intuitive than some of the `awk` one liners that stack overflow suggests to uniq without sort.

Is this use case still not interesting enough to be merged?

thp:
Yeah I think it's valid, but rename it `remove-duplicate-lines or something so that it's easier for non-Unix people and so that there's no confusion with the Unix `uniq` tool which works slightly differently.

Sveder:
@thp updated the name as per your suggestion.

lib/urlwatch/filters.py
```diff
@@ -809,6 +809,24 @@ def filter(self, data, subfilter):
...
+
+    __default_subfilter__ = 'separator'
+
+    def filter(self, data, subfilter):
+        separator = subfilter.get('separator', '\n')
+        data_lines = data.split(separator)
+        uniq_lines = []
+        for line in data_lines:
+            if line not in uniq_lines:
+                uniq_lines.append(line)
```
thp, line 825: Depending on how many lines there are filtered, it might be faster (at the cost of additional memory) to put the "seen lines" in a `set`, since `set` membership test is cheaper than list membership test.

thp: Since there's no ordered set in Python, you'd need to keep the elements around twice, but since the only use for the `uniq_lines` list is to join it, this can be turned into a generator function (inline or otherwise):

```python3
def get_unique_lines(data_lines):
    seen = set()
    for line in data_lines:
        if line not in seen:
            yield line
            seen.add(line)

return separator.join(get_unique_lines(data_lines))
```

Of course, `str.split()` also returns a list of strings, when it could in theory return an iterable, but [since the proposed solutions](https://stackoverflow.com/questions/3862010) are all kind of cumbersome, let's not go that rounte.

thp (changes requested):
See comments, rest looks good already. The optional `separator` subfilter option should also be documented.

Sveder:
@thp made the changes :)

docs/source/filters.rst
```diff
@@ -69,6 +69,8 @@ At the moment, the following filters are built-in:
 - **sha1sum**: Calculate the SHA-1 checksum of the content
 - **shellpipe**: Filter using a shell command
 - **sort**: Sort input items
+- **remove-duplicate-lines**: Remove duplicate lines (case sensitive). Optional `separator` filter
+  to control how input is split into lines, defaults to `\n'.
```
thp, line 73: Remove the `Optional `separator` filter to control how input is split into lines, defaults to `\n'.` part, as this should be in the details.

lib/urlwatch/filters.py
```diff
@@ -809,6 +809,27 @@ def filter(self, data, subfilter):
         return separator.join(sorted(data.split(separator), key=str.casefold, reverse=reverse))
 
 
+class RemoveDuplicateLinesFilter(FilterBase):
+    """Remove duplicate lines"""
+
+    __kind__ = 'remove-duplicate-lines'
+
```
thp, line 816: Add this here:

```
    __supported_subfilters__ = {
        'separator': 'Item separator (default: newline)',
    }
```

thp (changes requested):
See comments.

Sveder:
Changes made.

thp (approved):

thp:
Please mark as ready for review + update changelog + squash to a single commit and then we can merge this.

Sveder:
@thp done.

Issue #608, opened 2020-12-26
why is user_visible_url limited to url jobs?


kongomongo:
I think it would be suitable for a number of other jobs, most prominently a command job where you have to use spezialized tools to access an api, but still want the html report header to be a clickable url?

thp:
Fixed by merging #654.

PR #654, opened 2021-07-24
enable user_visible_url for command/shell jobs

kongomongo:
hi there. would really love to be able to do this for command/shell jobs. 

I have lots of jobs that are cumbersome to get via default url job because of special cookie setup or websocket connections etc. , so I put them in .sh scripts and run them as command jobs. However I cant make clickable url in my HtmlReporter.

The only thing I'm a bit puzzled about is `LOCATION_IS_URL` which should really be a function ?

### DIFF
--- a/docs/source/jobs.rst
+++ b/docs/source/jobs.rst
@@ -48,7 +48,6 @@ Job-specific optional keys:
 - ``ignore_http_error_codes``: List of HTTP errors to ignore (see :ref:`advanced_topics`)
 - ``ignore_timeout_errors``: Do not report errors when the timeout is hit
 - ``ignore_too_many_redirects``: Ignore redirect loops (see :ref:`advanced_topics`)
-- ``user_visible_url``: Different URL to show in reports (e.g. when watched URL is a REST API URL, and you want to show a webpage)
 
 (Note: ``url`` implies ``kind: url``)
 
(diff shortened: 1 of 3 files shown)


kongomongo:
hrm why is this failing

kongomongo:
alright, back on topic ;-)

Sveder:
This is a great idea - 👍

thp (changes requested):
Should this be part of `JobBase` instead, so that it can also be used for overriding the URL/location of any kind of job (and in this sense, rename it `user_visible_location`, since not all jobs have a URL as such).

kongomongo:
> Should this be part of `JobBase` instead, so that it can also be used for overriding the URL/location of any kind of job (and in this sense, rename it `user_visible_location`, since not all jobs have a URL as such).

I agree. Still, what do you suggest for the LOCATION_IS_URL variable? Should this be a function that just does a regex match if location is a URI?

thp:
> > Should this be part of `JobBase` instead, so that it can also be used for overriding the URL/location of any kind of job (and in this sense, rename it `user_visible_location`, since not all jobs have a URL as such).
> 
> I agree. Still, what do you suggest for the LOCATION_IS_URL variable? Should this be a function that just does a regex match if location is a URI?

Yeah, a regex match would most likely do the trick.

kongomongo:
@thp Can you have a look? I added a very simple regex.

Disclaimer: I have NOT tested the code at runtime.

thp (changes requested):
Whoops, I just realized now that `user_visible_url` as key was already released in version 2.22, so (even though I requested this earlier) we shouldn't rename the key (as to not break existing user's configurations). The rest of the code seems fine, and moving from url job to job base is also fine (also moving in the documentation is fine).

It's just so that existing 2.22 configurations with `user_visible_url` set should still work when upgrading to a newer release.

The rest is fine, and `location_is_url()` also looks sane from here..

kongomongo:
OK I changed it back. So now the only "breaking" change could be that if someone had user_visible_url set to "www.mystuff.com" then it is now not set as a location anymore now due to the regex only picking up true URI (http:// ftp://, etc). But I guess even in that case a HTML displaying device will make that a link anyways?

kongomongo:
just for reference... https://github.com/thp/urlwatch/issues/608

thp (approved):

thp:
Thanks!

PR #655, opened 2021-07-28
Add signature configuration

Sveder:
@thp what do you think?

I understand it might be a bit controversial as this is your signature, but my end-users are definitely not the audience for this message, so allowing to turn it off will make sense in some use cases.

### DIFF
--- a/docs/source/configuration.rst
+++ b/docs/source/configuration.rst
@@ -26,9 +26,13 @@ You can change this behavior in the ``display`` section of the configuration:
      new: true
      error: true
      unchanged: false
+     signature: true
 
 If you set ``unchanged`` to ``true``, urlwatch will always report all pages
 that are checked but have not changed.
+
+``signature``, which is on by default, is a small blurb about urlwatch and ways
+to support it.
 
 
 Filter changes are not applied for ``unchanged``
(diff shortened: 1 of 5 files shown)


thp (changes requested):
The uniq changes from #653 shouldn't be in here, let's continue discussion about that change over there.

I don't want to make it possible to remove the signature via a configuration option at the moment.

As you suspected, I don't really like the idea of turning off the signature, especially in this case where you apparently have "end users" (consider buying me a coffee if the tool is useful to you providing a service to your users!). You can of course do whatever you want on your machine locally (patch it out, change it), but for now I'd like to keep the upstream project to have this info.

The signature serves other useful purposes: It shows the version number (so a end user can easily see if it's the latest version without having to log into the server where urlwatch runs to check its version output), also it shows that it's actually the program running and sending e-mails (so it's not mistaken for spam and the user has a chance of figuring out where the mail is coming from and be reminded that they or somebody else set it up for them, and in turn figure out how to disable it should they want to) and the duration info is useful to see if watching sometimes takes long or if it's all fast and efficient (e.g. if it watches "2 URLs in 300 seconds", something might be off).

Sveder:
First of all - thanks for answering and your position is valid. 
I agree that the signature has some debugging information that might be relevant, but I believe that tools should let you configure them, and as I said, in my case both the personal and the debugging info is not interesting to the people reading the output.

Thanks again for your time!

PR #658, opened 2021-08-05
Add csv2text filter.

Sveder:


lib/urlwatch/filters.py
```diff
@@ -330,6 +331,46 @@ def filter(self, data, subfilter):
...
+            'format_message': 'A format string with the headers that will be outputted for each csv '
+                                                         'line (header will be lower-cased)',
+            'ignore_header': 'If your format string is number based, but the CSV has headers, '
+                                                       'this flag will force ignoring the header.',
+        }
+
+    __default_subfilter__ = 'format_message'
+
+    def filter(self, data, subfilter):
+        has_header = csv.Sniffer().has_header(data)
```
thp, line 349: Should it be possible to supply a subfilter, so that:

1. If the subfilter `has_header` is `True` or `False`, it overrides the `has_header`
2. If there is no `has_header` set in the subfilter, `csv.Sniffer()` can be used

lib/urlwatch/filters.py
```diff
@@ -330,6 +331,46 @@ def filter(self, data, subfilter):
...
+
+        reader = csv.reader(data.split('\n'))
+        data_list = list(reader)
+        header = None
+
+        if has_header:
+            header = data_list[0]
```
thp, line 356: `header = data_list.pop(0)` (then you don't need to slice the list below)

lib/urlwatch/filters.py
```diff
@@ -330,6 +331,46 @@ def filter(self, data, subfilter):
...
+            data_list = data_list[1:]
+
+        header = [i.lower() for i in header]
+        message = subfilter['format_message']
```
thp, line 360: Should there be a default / fallback for `format_message`?

How does the CSV output look like?

thp (changes requested):
See comments.

lib/urlwatch/filters.py
Sveder: Why is pop-ing better than slicing? 
IMHO, slicing looks better and more pythonic.

lib/urlwatch/filters.py
Sveder: It is hard to create a default, since the main ability of this is to modify the input (csv) to a pretty string that uses rows(/headers) as format data.
If I must think about something maybe the default with headers should be:
```
First header: {first_header_data}. Second header: {second_header_data}. ...
```

Without headers the csv itself is as much "default" as we can get (it doesn't make sense to omit or reorder the columns), so no reason to use this filter.

Do you think the no format string default with header is interesting enough to add?

lib/urlwatch/filters.py
Sveder: To make sure I understand, you are suggesting that if `has_header` is False we ignore sniffer result and just include the header as part of the data?
To me it seems like allowing someone to shoot themselves in the foot, but if you feel that this customization is needed, then sure, I'll add.

Sveder:
@thp would love a few clarifications. Thanks :)

lib/urlwatch/filters.py
thp: There's only one source code line that describes both, so nothing can go out of sync if one of the two lines are later edited. Also, no need to create a slice object or in the worst case a second list that nobody is going to touch again.

Also (not sure if it still applies to the latest Python 3 version): https://esaezgil.com/post/python_list_pop_slice/

lib/urlwatch/filters.py
thp: The thing is, the sniffer might be wrong, so it makes sense to force-disable it. If there's no way to disable it, and if it "guesses" wrong, the filter will be unusable for certain data.

lib/urlwatch/filters.py
thp: Yeah, I guess there's no point in having a default then. The docs should list a good example value for this, though.

Sveder:
@thp fixed lint and rebased.

thp:
Merged, thanks!

PR #663, opened 2021-08-27
Ensure code block starts with newline

ml-:
Code blocks are not visible if not started with a newline. This can be verified using the Github Preview function as well as in the current (2021-08-27T20:18+02:00) [documentation](https://urlwatch.readthedocs.io/en/latest/reporters.html#telegram)


thp (approved):
Well spotted, thanks!

PR #693, opened 2022-02-19
Add strip lines filter

Lukas0907:
This PR adds a strip lines filter which is similar to the existing `strip` filter. The difference is that it strips every line in the output and not only the beginning and the end of the output.


thp (approved):

thp:
Merged, thanks.

PR #696, opened 2022-03-14
renew github releases example

julianuu:
The old example ceased to work.

### DIFF
--- a/docs/source/filters.rst
+++ b/docs/source/filters.rst
@@ -403,13 +403,23 @@ project for the latest release version, to be notified of new releases:
 
 .. code:: yaml
 
-   url: https://github.com/thp/urlwatch/releases
+   url: https://github.com/tulir/gomuks/releases
    filter:
-     - xpath: '(//div[contains(@class,"release-timeline-tags")]//h4)[1]/a'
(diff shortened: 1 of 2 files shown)


thp (approved):
Thanks!

docs/source/filters.rst
```diff
@@ -403,13 +403,23 @@ project for the latest release version, to be notified of new releases:
 
 .. code:: yaml
 
-   url: https://github.com/thp/urlwatch/releases
+   url: https://github.com/tulir/gomuks/releases
```
neutric, line 406: @thp Was this change of `url` merged intentionally?

neutric: That makes sense.

thp: Yeah, i had to do a double-take too, but as @julianuu said, there are now two "unit tests" for those filters, so the URL needs to be unique. The unit test doesn't do HTTP requests, and just uses the local test data, so it will continue working as an example (for unit tests) even if the repo goes away.

julianuu: Sorry, it was not on purpose that the urls have the same length ;)

PR #698, opened 2022-03-16
test_filter: fix ResourceWarning: unclosed file

sbraz:
The warning was visible when running `pytest -Wdefault lib/urlwatch/tests/test_filters.py`.


thp (approved):
Thanks!

Issue #670, opened 2021-09-08
Proposals for "navigate": ability to set a custom user agent and to execute a javascript in the page context


maiux:
I like this project, if you agree I think this two features will be an improvement, since some websites knock down the request when they spot an "*/headless" user agent.

I hope it will be easy to set a custom user agent and evaluate javascript in the _get_content() function of browser module using page.setUserAgent and page.evaluate method exposed.

I think also it will be a great addition using playwright as an alternative browser interface, since playwright IMHO has a much smoother experience handling automation and has a native python API

thp:
Both suggestions sound fine to me, patches gladly accepted.

PR #700, opened 2022-04-07
Support change of useragent for Browser job

fversaci:
This should solve issue #670 

Please check/modify as you like.

- Added optional key "useragent"
- Using it to set the useragent for pyppeteer
- Added description of key in documentation
  (the option is still missing in the man pages though)


thp:
Merged, thanks!

Issue #76, opened 2016-06-25
Atom feed reporter


fstirlitz:
I'd love to have urlwatch output an Atom feed which I'd then be able to import into my feed reader.

Liferea supports reading feeds generated by a command-line program; it could work great together with urlwatch.

thp:
Closing as duplicate of https://github.com/thp/urlwatch/issues/76 -- whether we use RSS or Atom should just be an implementation detail, feed readers can usually handle both.

PR #709, opened 2022-05-27
WIP: Add RSS Reporter with feedgen

georgjaehnig:
Hello, because needing it myself, I've picked up the work on the RSS reporter (#553, #76, #53). I tried a very basic approach: Simply take the STDOUT output and wrap it into an RSS feed with 1 item.

Here's [https://jaehnig.org/urlwatch/rss.xml an example output].

If `urlwatch` now runs e.g. once a day, then in my feedreader I get this new item once per day. When using on online feed reader like Feedly, the old entries get saved by Feedly anyway.

Create different feeds for different jobs should also be possible via

    urlwatch 1 > 1.xml
    urlwatch 2 > 2.xml

I think this is very much "good enough", much better than no RSS feed at all. But of course I'm happy to polish this up if there are some low-hanging fruits. :)

### DIFF
--- a/lib/urlwatch/reporters.py
+++ b/lib/urlwatch/reporters.py
@@ -46,6 +46,8 @@ from .mailer import SendmailMailer
 from .util import TrackSubClasses, chunkstring
 from .xmpp import XMPP
 
+from feedgen.feed import FeedGenerator
+
 try:
     import chump
 except ImportError:
--- a/lib/urlwatch/reporters.py
+++ b/lib/urlwatch/reporters.py
@@ -385,6 +385,35 @@ class StdoutReporter(TextReporter):
                 print(line)
 
 
+class RssReporter(TextReporter):
+    """Print summary on stdout (the console) as RSS"""
+
+    __kind__ = 'rss'
+
+    def submit(self):
+
+        body = '\n'.join(super().submit())
+        if not body:
+            return
+        
+        # Text output within preformatted tag.
+        body = '<pre>' + body + '</pre>'
+
+        # Create feed.
+        fg = FeedGenerator()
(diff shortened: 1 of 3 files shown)


thp (changes requested):
Wonder if it could be better to configure the output file in the settings instead of writing to stdout? You could e.g. allow format strings to write per-feed files:

```
    output_filename: 'everything.rss'
```

vs

```
    output_filename: 'feed_{name}.rss'
```

(where `{name}` would be taken from the "name" field in the job list)

If you use `output_filename.format(**job)` (and `job` is the YAML dict) then you could even have custom keys and insert those into the filenames.

One important thing there would of course be to "collect" all jobs that map to the same filename and generate the feed once.

thp:
@georgjaehnig Did you have time to look at the PR review comment?

georgjaehnig:
Hi! Yes, I've read it back then – but in the meantime I was more concerned with the fact that (at least) Feedly does not recognize new items: Even if the content differed, it did not show a new item. Still have to work on this.

About the idea with configuring the output file: I'm not sure if I understand the benefit. Can I achieve something more with this than just by taking `STDOUT` and redirecting it with `>` to the file of my desire?

georgjaehnig:
> You can have multiple feeds (e.g. "ebay.rss", "news.rss", ...)

But I could achieve that already with `STDOUT`, like this, no?

    urlwatch ebay > ebay.rss
    urlwatch news > news.rss

thp:
@Jamstah What do you think of this PR? Do you think we should clean it up and get it merged or just close unmerged for now?

georgjaehnig:
I'd be also happy if this moved on. :) Sorry for leaving this dormant. But I can help now again.

PR #707, opened 2022-05-25
Documentation: how to ignore whitespace changes

mentalstring:
As per #706.

Added to Advanced Topics, under word-based differences, just because it's somewhat related. Happy to move it elsewhere if more appropriate.

PR #712, opened 2022-05-30
docs: group diff-related topics and tweak wording

neutric:
This puts a bit more order in the already existing topics by grouping all diff-related topics. Also, wording from https://github.com/thp/urlwatch/pull/707 is tweaked to be more specific.


thp (approved):
Thanks :)

PR #715, opened 2022-07-29
The dependency is html2text, not pyhtml2text.

JulienPalard:
It could be misleading as people may install https://github.com/carsonip/pyhtml2text instead.


thp (approved):

thp:
Thanks!

PR #731, opened 2022-12-06
shrink container image from 1g to 150m

sedlund:
The original container build weighed in at 1.03g.  I switched it to the latest stable python 3.11.0 image based on Alpine Linux and this brings it down to 151m.  I tested with the latest stable with their Debian bullseye slim image and it comes to 228m.

I also converted the container to run as a user instead of root and use pip to install from local.

Let me know what you think.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,16 +1,22 @@
-FROM python:3.10.0
+# https://hub.docker.com/_/python
+FROM python:3.11.0-alpine3.17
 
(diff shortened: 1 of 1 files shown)


sedlund:
Wanted to tag @chenrui333 and @agix for comments

agix (approved):
Sounds good to me

sedlund:
Hi @thp, I think I completed the needful, if not let me know.  Thanks.

thp:
Merged, thanks!

Issue #737, opened 2022-12-19
UnicodeDecodeError: 'charmap' codec can't decode byte


yuis-ice:
When I have the following urls.yaml, it succeeds, 

```
kind: url
name: rakuten card promo
url: https://www.rakuten-card.co.jp/campaign/add-card/
filter: 
  - css: "section[id='rule_detail']"
  - html2text:
    method: re
  - grep: "2,000"
```

```
> urlwatch --urls urls.yaml --test-filter 1
                2枚目の楽天カードを作成＆利用特典…2,000ポイント（期間限定ポイント）
```

but with the following urls.yaml, where I have a utf-8 text content on it, it gets an error.

```
kind: url
name: rakuten card promo
url: https://www.rakuten-card.co.jp/campaign/add-card/
filter: 
  - css: "section[id='rule_detail']"
  - html2text:
    method: re
  - grep: "2,000ポイント"
```

```
> urlwatch --urls urls.yaml --test-filter 1
Traceback (most recent call last):
  File "C:\Python37\lib\runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "C:\Python37\lib\runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "C:\pg\urlwatch_dev\venv\Scripts\urlwatch.exe\__main__.py", line 7, in <module>
  File "c:\pg\urlwatch_dev\venv\lib\site-packages\urlwatch\cli.py", line 108, in main
    urlwatch = Urlwatch(command_config, config_storage, cache_storage, urls_storage)
  File "c:\pg\urlwatch_dev\venv\lib\site-packages\urlwatch\main.py", line 66, in __init__
    self.load_jobs()
  File "c:\pg\urlwatch_dev\venv\lib\site-packages\urlwatch\main.py", line 85, in load_jobs
    jobs = self.urls_storage.load_secure()
  File "c:\pg\urlwatch_dev\venv\lib\site-packages\urlwatch\storage.py", line 316, in load_secure
    jobs = self.load()
  File "c:\pg\urlwatch_dev\venv\lib\site-packages\urlwatch\storage.py", line 419, in load
    return self._parse(fp)
  File "c:\pg\urlwatch_dev\venv\lib\site-packages\urlwatch\storage.py", line 385, in _parse
    jobs = [JobBase.unserialize(job) for job in yaml.load_all(fp, Loader=yaml.SafeLoader)
[...]

PR #738, opened 2022-12-19
fixes UnicodeDecodeError for non English urls.yaml

yuis-ice:


yuis-ice:
fixes #737

thp (changes requested):
While I get the intention, just catching any `UnicodeDecodeError` is wrong, as it will also fail if the system tries to decode as UTF-8 and then decode twice.

The problem is that Python 3 on Windows 10 doesn't use UTF-8 by default. If you set the environment variable `PYTHONUTF8=1` ([documentation](https://dev.to/methane/python-use-utf-8-mode-on-windows-212i), [PEP-540](https://peps.python.org/pep-0540/)), then it should work properly.

thp:
> While I get the intention, just catching any `UnicodeDecodeError` is wrong, as it will also fail if the system tries to decode as UTF-8 and then decode twice.
> 
> The problem is that Python 3 on Windows 10 doesn't use UTF-8 by default. If you set the environment variable `PYTHONUTF8=1` ([documentation](https://dev.to/methane/python-use-utf-8-mode-on-windows-212i), [PEP-540](https://peps.python.org/pep-0540/)), then it should work properly.

..and as documented in PEP 540, this requires Python 3.7 and newer.

thp:
However, it might make sense to add this somewhere to the documentation so that it's easier to find for users.

yuis-ice:
I understand that. I don't have a good idea for a better solution instead of catching UnicodeDecodeError for now.

thp:
@yuis-ice Did you test if setting `PYTHONUTF8=1` locally in your environment will fix the issue even without this code change?

yuis-ice:
Yes it works. 

```
> $env:PYTHONUTF8 = '1'

> urlwatch --urls urls.yaml --test-filter 1
                2枚目の楽天カードを作成＆利用特典…2,000ポイント（期間限定ポイント）
```

thp:
UTF-8 Mode for Windows is now documented here: 3341d880ee7fc7b899abc788570fd0567473d365

PR #739, opened 2022-12-19
Implements command to change a job location.

### LINKED ISSUE #300: Allow changing URLs/commands while keeping change history
Sometimes I need to tweak the url or command for a given entry, without materially changing the content that gets fetched. But since urlwatch generates GUIDs by hashing the url or command string itself, this means the changed entry always gets treated as "new".

What about using the "name" field for identification instead? It would be a breaking change for people who currently have duplicate names, but I don't know how common that is.

trevorshannon:
This PR adds a new command line option `--change-location` to change the location (i.e. `url`, `command`, or `navigate`) of a job while preserving history for that job.  The tool ensures that the new location string is unique within the list of jobs prior to changing anything.

Note that in the (admittedly contrived) scenario of having abandoned entries in a redis db that use the user-provided new location (perhaps from old runs that were never garbage-collected) then those abandoned entries will be clobbered by the `--change-location` option.  I think that should be ok, but I'm sure I could find a way to preserve those entries if needed.

I did some manual testing to verify functionality.

Save the starting version of a simple urls YAML:
```
➜  urlwatch git:(feat/update-url)  cp ~/urls.yaml ~/urls_init.yaml
➜  urlwatch git:(feat/update-url) cat ~/urls_init.yaml 
command: date
kind: shell
---
kind: url
url: https://www.example.org
---
kind: url
url: https://www.wikipedia.org
user_visible_url: https://en.wikipedia.org
---
kind: browser
navigate: https://www.google.com
```

Change location of all 4 jobs:
```
➜  urlwatch git:(feat/update-url)  urlwatch --urls ~/urls.yaml --change-location 1 "date -v1m"                                
Moving location of <shell command='date'> to "date -v1m"
Saving updated list to '/Users/trevorshannon/urls.yaml'
➜  urlwatch git:(feat/update-url)  urlwatch --urls ~/urls.yaml --change-location https://www.example.org https://www.example.net
Moving location of <url url='https://www.example.org'> to "https://www.example.net"
Saving updated list to '/Users/trevorshannon/urls.yaml'
➜  urlwatch git:(feat/update-url)  urlwatch --urls ~/urls.yaml --change-location 3 https://es.wikipedia.org                     
Moving location of <url url='https://www.wikipedia.org' user_visible_url='https://en.wikipedia.org'> to "https://es.wikipedia.org"
Saving updated list to '/Users/trevorshannon/urls.yaml'
➜  urlwatch git:(feat/update-url)  urlwatch --urls ~/urls.yaml --change-location 4 https://www.google.co.uk
Moving location of <browser navigate='https://www.google.com'> to "https://www.google.co.uk"
Saving updated list to '/Users/trevorshannon/urls.yaml'
```

Compare the updated urls YAML to the initial reference version. Appropriate key values are updated according to job type:
```
➜  urlwatch git:(feat/update-url)  diff ~/urls_init.yaml ~/urls.yaml   
1c1
< command: date
---
> command: date -v1m
5c5
< url: https://www.example.org
---
> url: https://www.example.net
9c9
< user_visible_url: https://en.wikipedia.org
---
> user_visible_url: https://es.wikipedia.org
12c12
< navigate: https://www.google.com
---
> navigate: https://www.google.co.uk
```

Existing history is moved with --change-location:
```
➜  urlwatch git:(feat/update-url) urlwatch --urls ~/urls.yaml --dump-history 1
==============================
2022-12-20 13:08
------------------------------
Tue Dec 20 13:08:59 EST 2022

============================== 

==============================
2022-12-20 13:18
------------------------------
Tue Dec 20 13:18:54 EST 2022

============================== 

2 historic snapshot(s) available
➜  urlwatch git:(feat/update-url) urlwatch --urls ~/urls.yaml --change-location date "date -v1m"
Moving location of <shell command='date'> to "date -v1m"
Moved 2 snapshots of "date" to "date -v1m"
Saving updated list to '/Users/trevorshannon/urls.yaml'
➜  urlwatch git:(feat/update-url) urlwatch --urls ~/urls.yaml                                   
===========================================================================
01. CHANGED: date -v1m
===========================================================================

CHANGED: date -v1m
--- @	Tue, 20 Dec 2022 13:18:54 -0500
+++ @	Tue, 20 Dec 2022 13:20:33 -0500
@@ -1 +1 @@
-Tue Dec 20 13:18:54 EST 2022
+Thu Jan 20 13:20:33 EST 2022

➜  urlwatch git:(feat/update-url) urlwatch --urls ~/urls.yaml --dump-history 1                  
==============================
2022-12-20 13:08
------------------------------
Tue Dec 20 13:08:59 EST 2022

============================== 

==============================
2022-12-20 13:18
------------------------------
Tue Dec 20 13:18:54 EST 2022

============================== 

==============================
2022-12-20 13:20
------------------------------
Thu Jan 20 13:20:33 EST 2022

============================== 

3 historic snapshot(s) available
```

I confirmed the same behavior with `--cache redis://127.0.0.1:6379`

Fixes #300

### DIFF
--- a/lib/urlwatch/command.py
+++ b/lib/urlwatch/command.py
@@ -212,6 +212,31 @@ class UrlwatchCommand:
             print('Adding %r' % (job,))
             self.urlwatcher.jobs.append(job)
 
+        if self.urlwatch_config.change_location is not None:
+            new_loc = self.urlwatch_config.change_location[1]
+            # Ensure the user isn't overwriting an existing job with the change.
+            if new_loc in [j.get_location() for j in self.urlwatcher.jobs]:
+                print('The new location "%s" already exists for a job. Use that'
+                      ' existing job or choose a different value.' % new_loc)
+                save = False
+            else:
+                job = self._find_job(self.urlwatch_config.change_location[0])
+                if job is not None:
+                    # Update the job's location (which will also update the 
+                    # guid) and move any history in the cache over to the job's 
+                    # updated guid.
+                    print('Moving location of %s to "%s"' % (job, new_loc))
+                    old_guid = job.get_guid()
+                    old_loc = job.get_location()
+                    job.set_location(new_loc)
+                    num_moved = self.urlwatcher.cache_storage.move(
+                        old_guid, job.get_guid())
+                    if num_moved:
+                        print('Moved %d snapshots of "%s" to "%s"' % (num_moved, old_loc, new_loc))
+                else:
+                    print('Not found: %r' % self.urlwatch_config.change_location[0])
+                    save = False
+
         if save:
             self.urlwatcher.urls_storage.save(self.urlwatcher.jobs)
 
@@ -235,7 +260,9 @@ class UrlwatchCommand:
             sys.exit(self.dump_history(self.urlwatch_config.dump_history))
         if self.urlwatch_config.list:
             sys.exit(self.list_urls())
-        if self.urlwatch_config.add is not None or self.urlwatch_config.delete is not None:
+        if (self.urlwatch_config.add is not None or 
+            self.urlwatch_config.delete is not None or 
+            self.urlwatch_config.change_location is not None):
             sys.exit(self.modify_urls())
 
     def check_edit_config(self):
--- a/lib/urlwatch/command.py
+++ b/lib/urlwatch/command.py
@@ -216,8 +216,8 @@ class UrlwatchCommand:
             new_loc = self.urlwatch_config.change_location[1]
             # Ensure the user isn't overwriting an existing job with the change.
             if new_loc in (j.get_location() for j in self.urlwatcher.jobs):
-                print('The new location "%s" already exists for a job. Use that '
-                      'existing job or choose a different value.' % new_loc)
+                print(f'The new location "{new_loc}" already exists for a job. Use that '
+                      'existing job or choose a different value.')
                 save = False
             else:
                 job = self._find_job(self.urlwatch_config.change_location[0])
@@ -225,14 +225,14 @@ class UrlwatchCommand:
                     # Update the job's location (which will also update the
                     # guid) and move any history in the cache over to the job's
                     # updated guid.
-                    print('Moving location of %r to "%s"' % (job, new_loc))
+                    print(f'Moving location of {job!r} to "{new_loc}"')
                     old_guid = job.get_guid()
                     old_loc = job.get_location()
                     job.set_location(new_loc)
                     num_moved = self.urlwatcher.cache_storage.move(
                         old_guid, job.get_guid())
                     if num_moved:
-                        print('Moved %d snapshots of "%s" to "%s"' % (num_moved, old_loc, new_loc))
+                        print(f'Moved {num_moved} snapshots of "{old_loc}" to "{new_loc}"')
                 else:
                     print('Not found: %s' % self.urlwatch_config.change_location[0])
                     save = False
--- a/lib/urlwatch/command.py
+++ b/lib/urlwatch/command.py
@@ -228,7 +228,7 @@ class UrlwatchCommand:
                     print(f'Moving location of {job!r} to "{new_loc}"')
                     old_guid = job.get_guid()
                     old_loc = job.get_location()
-                    job.set_location(new_loc)
+                    job.set_base_location(new_loc)
                     num_moved = self.urlwatcher.cache_storage.move(
                         old_guid, job.get_guid())
                     if num_moved:
--- a/lib/urlwatch/command.py
+++ b/lib/urlwatch/command.py
@@ -216,8 +216,8 @@ class UrlwatchCommand:
             new_loc = self.urlwatch_config.change_location[1]
             # Ensure the user isn't overwriting an existing job with the change.
             if new_loc in (j.get_location() for j in self.urlwatcher.jobs):
-                print(f'The new location "{new_loc}" already exists for a job. Use that '
-                      'existing job or choose a different value.')
+                print(f'The new location "{new_loc}" already exists for a job. '
+                      'Delete the existing job or choose a different value.')
                 save = False
             else:
                 job = self._find_job(self.urlwatch_config.change_location[0])
@@ -234,7 +234,7 @@ class UrlwatchCommand:
                     if num_moved:
                         print(f'Moved {num_moved} snapshots of "{old_loc}" to "{new_loc}"')
                 else:
-                    print('Not found: %s' % self.urlwatch_config.change_location[0])
+                    print(f'Not found: {self.urlwatch_config.change_location[0]}')
                     save = False
 
         if save:
--- a/lib/urlwatch/command.py
+++ b/lib/urlwatch/command.py
@@ -215,17 +215,17 @@ class UrlwatchCommand:
         if self.urlwatch_config.change_location is not None:
             new_loc = self.urlwatch_config.change_location[1]
             # Ensure the user isn't overwriting an existing job with the change.
-            if new_loc in [j.get_location() for j in self.urlwatcher.jobs]:
-                print('The new location "%s" already exists for a job. Use that'
-                      ' existing job or choose a different value.' % new_loc)
+            if new_loc in (j.get_location() for j in self.urlwatcher.jobs):
+                print('The new location "%s" already exists for a job. Use that '
+                      'existing job or choose a different value.' % new_loc)
                 save = False
             else:
                 job = self._find_job(self.urlwatch_config.change_location[0])
                 if job is not None:
-                    # Update the job's location (which will also update the 
-                    # guid) and move any history in the cache over to the job's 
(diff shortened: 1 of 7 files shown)


lib/urlwatch/jobs.py
```diff
@@ -259,6 +268,12 @@ class UrlJob(Job):
     def get_location(self):
         return self.user_visible_url or self.url
 
+    def set_location(self, location):
+        if self.user_visible_url:
+            self.user_visible_url = location
+        else:
+            self.url = location
```
thp, line 275: This means that if a job has a `user_visible_url` set, the "real" URL isn't updated? Not sure if that's what a user would expect. Seems like user visible URL is something that's only for "cosmetics" (display), and the `--change-location` feature should always work on the "real" / "watched" URL? (same comment applies to the other `set_location()` functions).

lib/urlwatch/command.py
```diff
@@ -212,6 +212,31 @@ def modify_urls(self):
             print('Adding %r' % (job,))
             self.urlwatcher.jobs.append(job)
 
+        if self.urlwatch_config.change_location is not None:
+            new_loc = self.urlwatch_config.change_location[1]
+            # Ensure the user isn't overwriting an existing job with the change.
+            if new_loc in (j.get_location() for j in self.urlwatcher.jobs):
+                print('The new location "%s" already exists for a job. Use that '
+                      'existing job or choose a different value.' % new_loc)
```
thp, line 220: Use [f-strings](https://docs.python.org/3/tutorial/inputoutput.html#formatted-string-literals) instead of `%`-formatting.

lib/urlwatch/command.py
```diff
@@ -212,6 +212,31 @@ def modify_urls(self):
...
+                    # guid) and move any history in the cache over to the job's
+                    # updated guid.
+                    print('Moving location of %r to "%s"' % (job, new_loc))
+                    old_guid = job.get_guid()
+                    old_loc = job.get_location()
+                    job.set_location(new_loc)
+                    num_moved = self.urlwatcher.cache_storage.move(
+                        old_guid, job.get_guid())
+                    if num_moved:
+                        print('Moved %d snapshots of "%s" to "%s"' % (num_moved, old_loc, new_loc))
```
thp, line 235: f-string here as well.

lib/urlwatch/command.py
```diff
@@ -212,6 +212,31 @@ def modify_urls(self):
...
+                else:
+                    print('Not found: %s' % self.urlwatch_config.change_location[0])
```
thp, line 237: f-string, too.

thp (changes requested):
See comments.

lib/urlwatch/jobs.py
trevorshannon: Yes, this threw me for a loop a bit.  I expected `user_visible_url` to simply be cosmetic, but it seems as though if `user_visible_url` is set, _that_ gets [used](https://github.com/thp/urlwatch/blob/c53876b02597d76e59ca2d0a15a7eb5ef0573fca/lib/urlwatch/jobs.py#L260) to compute the GUID, not the "true" URL.  I'm not too sure why the `user_visible_url` is used as the `location` to compute the GUID, but since it is, it seemed like that should be the value modified when editing the `location`

Hmmm, though now that I think about it more, if we leave the `user_visible_url` untouched and only ever modify the true URL then the job history _will_ be retained since the GUID will also be untouched in that situation.  The only downside is that for jobs with `user_visible_url` set, `get_location()` will not return the string passed in with a previous call to `set_location()`.  I suppose `set_location()` could be renamed though I'm not sure of a name that would make sense across all three job types.

thp: The thing with `user_visible_url` is that it was added as a "cosmetic" thing. Probably it accidentally got used as base for the GUID.

Here's one idea:

- Introduce an `original_guid` field that's empty by default and gets set by this migration
- If `user_visible_url` is set, but `original_guid` is not, it's used as the basis (like now, for backwards compatibility)
- If `original_guid` is set, it always "wins" over any other setting

This way, the migration only needs to set `original_guid` and it will work properly, and there's not even the need to rewrite the history, and existing behaviour isn't changed as long as original_guid is not set.

trevorshannon: Yes, that would work though I'm not a big fan of writing the GUID in the config file.  I know I'd be surprised to see a long random string show up in my config.  I have also benefited in the past from being able to manually compute a GUID from a url string in tests and one-off scripts.  Storing an original GUID based on some previous value of the url string would make it harder to do that conversion manually since it sort of "overrides" the GUID computation.

I think this achieves the same result without "saving" the GUID:

- Rename `set_location()` to `set_base_location()` and have it only ever set `self.url`, `self.command`, or `self.navigate` according to the job type. `user_visible_url` is untouched
- `get_location()` remains unchanged for backwards compatibility

That should result in the following behavior:

- If a job has `user_visible_url` set and this migration is executed, the `(url|command|navigate)` value is updated as one would expect.  The GUID never changes since `user_visible_url` is still the basis. History is retained.
- If a job does _not_ have `user_visible_url` set and this migration is executed, the `(url|command|navigate)` value is updated as one would expect and the GUID changes too.  The history is edited to match and is therefore retained.

What do you think?

thp: That sounds sensible, let's go with that. Also it should be backwards compatible and not be a behavior change compared to what was before (people adding a `user_visible_url` after the fact would still get their history lost, except if it's used as a "watchable URL has changed, but we want to keep history, so we'll put the old user visible URL there", for which this feature now exists).

thp (changes requested):
Let's go with your suggestion from [this comment](https://github.com/thp/urlwatch/pull/739/files#r1063191217).

lib/urlwatch/jobs.py
trevorshannon: Cool, I just implemented that.  Thanks for the thoughtful discussion.

lib/urlwatch/command.py
```diff
@@ -212,6 +212,31 @@ def modify_urls(self):
...
+                    print(f'Moving location of {job!r} to "{new_loc}"')
+                    old_guid = job.get_guid()
+                    old_loc = job.get_location()
+                    job.set_base_location(new_loc)
+                    num_moved = self.urlwatcher.cache_storage.move(
+                        old_guid, job.get_guid())
+                    if num_moved:
+                        print(f'Moved {num_moved} snapshots of "{old_loc}" to "{new_loc}"')
+                else:
+                    print('Not found: %s' % self.urlwatch_config.change_location[0])
```
thp, line 237: Please change this line to:

```
print(f'Not found: {self.urlwatch_config.change_location[0]}')
```

lib/urlwatch/command.py
```diff
@@ -212,6 +212,31 @@ def modify_urls(self):
             print('Adding %r' % (job,))
             self.urlwatcher.jobs.append(job)
 
+        if self.urlwatch_config.change_location is not None:
+            new_loc = self.urlwatch_config.change_location[1]
+            # Ensure the user isn't overwriting an existing job with the change.
+            if new_loc in (j.get_location() for j in self.urlwatcher.jobs):
+                print(f'The new location "{new_loc}" already exists for a job. Use that '
+                      'existing job or choose a different value.')
```
thp, line 220: Style nitpick:
```
                print(f'The new location "{new_loc}" already exists for a job. '
                      'Use that existing job or choose a different value.')
```

Semantic feedback:
What does "Use that existing job" mean in this context? Apart from accidental use (in which case the user sees the error message and goes on to correct their mistake), if people WANT to merge two different histories, maybe we should somehow allow this (and choose one or the other in the job list, maybe the "target" job gets overwritten using `--overwrite` or `--force` or something?). I guess the workaround now would be to just delete the target job from the jobs list manually (keeping historic entries in the database) and then do the rename? If this is good enough, then I suggest a wording change:

```
                print(f'The new location "{new_loc}" already exists for a job. '
                      'Delete the existing job or choose a different value.')
```

thp (changes requested):
Just two small style/wording comments. Can you also add an entry to `CHANGELOG.md` in the "Added" category describing the new feature? (+ PR number + your name as contributor, like the other examples already in the file)

After that, I think this is finally ready to merge. Thanks for pushing this PR forward, I think it turned out quite nice.

lib/urlwatch/command.py
trevorshannon: oops, good catch.

lib/urlwatch/command.py
trevorshannon: I was thinking that if somebody was updating a URL to watch example.com but they already have a job that watches example.com, then they can just use the existing example.com job to get whatever data they need.  Hence "Use", though I like "Delete" better.

trevorshannon:
Ok, all set.  Thanks for the feedback!

thp (approved):

thp:
Merged, thanks!

PR #746, opened 2022-12-21
- Fix decoding of strings for Redis Cache

### LINKED ISSUE #743: --gc-cache does not delete unused keys from redis db
When using a redis cache, the `--gc-cache` option successfully removes old entries, but does not delete old keys (despite claiming to do so).  This seems to be due to a string/bytes type issue somewhere around [here](https://github.com/thp/urlwatch/blob/25269fb6a788316b4f89b931cb40d763f7467230/lib/urlwatch/storage.py#L632).

When deleting unused keys, the system attempts to remove e.g. `"guid:b'27687e4f702f8547973bc60451b4f1539ee3d428'"` instead of `"guid:27687e4f702f8547973bc60451b4f1539ee3d428"`

scottmac:
Fixes #743, the strings weren't being decoded prior to adding them the list of known guids


thp (approved):

thp:
Thanks!

PR #752, opened 2023-04-05
add missing required dependency bs4

monperrus:


thp:
As mentioned in [the docs](https://urlwatch.readthedocs.io/en/latest/dependencies.html) ([source](https://github.com/thp/urlwatch/blob/master/docs/source/dependencies.rst)), `bs4` is optional. What was missing in the documentation is that `bs4` is also needed for `html2text` with `method: bs4`, this has been fixed in 56c8b56b8525fff9aa53addd466e17703ac84604. Thanks!

PR #756, opened 2023-04-17
Dockerfile: OPT_PYPKGS should be build arg, not env

PeterDaveHello:
build arg can be passed when building Docker image, and won't left after image build


thp (approved):

thp:
Merged, thanks!

PR #759, opened 2023-04-26
Update asyncio coroutine syntax to Python 3.11+

fasterit:
Cf. https://bugs.python.org/issue43216

Resolves the error when using the browser job type in Python 3.11.x:

  File "/[..]/lib/python3.11/site-packages/urlwatch/handler.py", line 68, in __enter__
    self.job.main_thread_enter()
  File "/[..]/lib/python3.11/site-packages/urlwatch/jobs.py", line 418, in main_thread_enter
    from .browser import BrowserContext
  File "/[..]/lib/python3.11/site-packages/urlwatch/browser.py", line 50
    browser = yield from pyppeteer.launch()
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
SyntaxError: 'yield from' inside async function

Also adds --no-sandbox as a default to the pyppeteer chromium invocation.

/DLange

### DIFF
--- a/lib/urlwatch/browser.py
+++ b/lib/urlwatch/browser.py
@@ -46,25 +46,23 @@ class BrowserLoop(object):
         self._loop_thread = threading.Thread(target=self._event_loop.run_forever)
         self._loop_thread.start()
 
-    @asyncio.coroutine
-    def _launch_browser(self):
-        browser = yield from pyppeteer.launch()
-        for p in (yield from browser.pages()):
-            yield from p.close()
+    async def _launch_browser(self):
+        browser = await pyppeteer.launch(headless=True, args=['--no-sandbox'])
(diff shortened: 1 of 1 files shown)


thp:
Thanks! However, this might become obsolete with #761?

fasterit:
If you want to fully migrate to playwright, yes.
That PR did not exist three days ago :-)
Greets, /DLange

thp:
I'm still going to merge this, since we require Python >= 3.7, and people who don't want to migrate to Playwright might find this useful. After merging this, I'm looking at the Playwright change next.

thp:
This is now in urlwatch 2.27, and urlwatch 2.28 migrates to Playwright, so users wanting to use pyppeteer for some reason can stick to 2.27 for now, and use Python 3.11.

Issue #751, opened 2023-03-12
Consider switching to playwright-python from pyppeteer


polyzen:
https://github.com/pyppeteer/pyppeteer#attention-this-repo-is-unmaintained-and-has-been-outside-of-minor-changes-for-a-long-time-please-consider-playwright-python-as-an-alternative

https://playwright.dev/python/

thp:
Fixed in urlwatch 2.28 via #761.

PR #761, opened 2023-04-28
Migrate from pyppeteer to playwright-python

paulsattlegger:
Having previously used a workaround to get the browser job to work on ARM devices, I have spent some time searching for a better approach. Rather than relying on the unmaintained dependency Pyppeteer, I would recommend switching to [Playwright](https://github.com/microsoft/playwright-python). Playwright is actively maintained by Microsoft (including the Python bindings), and provides a clean API that simplifies the browser job code. I have also updated the documentation related to the browser job. If there is anything else that needs to be done, I am happy to help.


polyzen:
Resolves #751

thp:
Thanks, merged and released in urlwatch 2.28 (with some [backwards compatibility and docs changes](https://github.com/thp/urlwatch/commit/2bdf79ea799ba5e4cdf7f58a30763b2c794887f1)).

PR #766, opened 2023-08-21
add more details on how to tweak javascript output

julianuu:
I don't know much about javascript or `playwright`, so at least for me it would have been helpful to have some more information here. For example I have been trying to watch a website which does not load properly unless I set the user agent to some common one, and set `wait_until` to `idle`.


docs/source/advanced.rst
```diff
@@ -319,13 +319,16 @@ To set the ``lynx`` output width to 400 characters, use this filter setup:
          width: 400
 
 
-Configuring how long browser jobs wait for pages to load
+Tweaking JavaScript output
 --------------------------------------------------------
```
thp, line 323: I think the title of that section was better before

docs/source/advanced.rst
```diff
@@ -319,13 +319,16 @@ To set the ``lynx`` output width to 400 characters, use this filter setup:
...
 
-For browser jobs, you can configure how long the headless browser will wait
+Browser jobs might sometimes fail to produce the desired results when loading websites that contain JavaScript.
+If JavaScript refuses to run you may try setting ``useragent`` to the user agent of the current version of a commonly used web browser.
```
thp, line 326: Do JavaScript libraries check for user-agent often?

docs/source/advanced.rst
```diff
@@ -319,13 +319,16 @@ To set the ``lynx`` output width to 400 characters, use this filter setup:
...
+You can configure how long the headless browser will wait
 before a page is considered loaded by using the ``wait_until`` option.
+This is useful for example if the job appears to be using the content of the website before the script finished running.
```
thp, line 329: This line seems to be redundant / does not add much to the description?

thp (changes requested):
See comments.

docs/source/advanced.rst
julianuu: I have no idea in general. But the first and only site with javascript that I am tracking so far, dhl tracking https://www.dhl.de/de/privatkunden/dhl-sendungsverfolgung.html, refuses to output tracking details with the default user agent.

thp:
Closing this for now.

PR #776, opened 2023-11-16
SlackReporter with preformatted rich text

vimagick:
```bash
$ urlwatch --test-reporter slack
```

## Slack will receive a message:

```
===========================================================================
01. NEW: Newly Added
02. CHANGED: Something Changed
03. ERROR: Error Reporting
===========================================================================

---------------------------------------------------------------------------
NEW: Newly Added ( http://example.com/new )
---------------------------------------------------------------------------


---------------------------------------------------------------------------
CHANGED: Something Changed ( http://example.com/changed )
---------------------------------------------------------------------------
--- @	Thu, 16 Nov 2023 13:42:21 +0000
+++ @	Thu, 16 Nov 2023 13:42:21 +0000
@@ -1,5 +1,5 @@
 
         Unchanged Line
-        Previous Content
+        Updated Content
         Another Unchanged Line
         
---------------------------------------------------------------------------


---------------------------------------------------------------------------
ERROR: Error Reporting ( http://example.com/error )
---------------------------------------------------------------------------
Traceback (most recent call last):
  File "/usr/lib/python3.10/site-packages/urlwatch/command.py", line 314, in set_error
    raise ValueError(message)
ValueError: Oh Noes!
---------------------------------------------------------------------------
```

Slack API Doc: <https://api.slack.com/reference/block-kit/blocks#rich_text>

PR #780, opened 2023-12-05
SlackReporter with preformatted rich text

vimagick:
Hi, I've updated my code ( #776 )

Three things:

- [x] Have you checked whether or not Mattermost supports Slack's rich text feature?

       Mattermost still use the original `{"text": text}`

- [x] How about users who want to opt out of this new behavior and have the old (non-rich-text) behavior? Should there be an option? What should be the default?

      Enabled only when `rich_text: true`

- [x] Please update CHANGELOG.md with information about this change

      Yes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ The format mostly follows [Keep a Changelog](http://keepachangelog.com/en/1.0.0/
 
 ## UNRELEASED
 
+### Added
+
+- Slack Reporter: `rich_text` config option for preformatted rich text
+
 ### Changed
 
 - Remove EOL'd Python 3.7 (new minimum requirement is Python 3.8), add Python 3.12 testing
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ The format mostly follows [Keep a Changelog](http://keepachangelog.com/en/1.0.0/
 
 ### Added
 
-- Slack Reporter: `rich_text` config option for preformatted rich text
+- Slack Reporter: `rich_text` config option for preformatted rich text (#780, by vimagick)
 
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -6,6 +6,10 @@ The format mostly follows [Keep a Changelog](http://keepachangelog.com/en/1.0.0/
 
 ## UNRELEASED
 
+### Added
+
+- Slack Reporter: `rich_text` config option for preformatted rich text
```
thp, line 11:
```suggestion
- Slack Reporter: `rich_text` config option for preformatted rich text (#780, by vimagick)
```

thp (changes requested):
You don't need to create a new PR when you update the PR. In fact, it makes reviewing a little more difficult. You can just update the PR. If you want to rewrite Git history, you can force-push to your remote branch and the PR will be updated.

Some more comments:

- [x] In the `CHANGELOG.md`, feel free to link the PR + add your credits
- [ ] Can you add `'rich_text': False,` to `lib/urlwatch/storage.py` in the `'slack':` section, so that it's easily discoverable?
- [ ] Can you add documentation for this to the `Slack` section in `docs/source/reporters.rst`

thp:
@vimagick Did you have time to look into these?

> * [ ]  Can you add `'rich_text': False,` to `lib/urlwatch/storage.py` in the `'slack':` section, so that it's easily discoverable?
> * [ ]  Can you add documentation for this to the `Slack` section in `docs/source/reporters.rst`

thp:
@vimagick Needs rebasing, and there's still the two items from above. Thanks!

thp:
Closing after no activity. Feel free to reopen and update the PR.

PR #785, opened 2024-01-17
Add option to disable individual jobs

### LINKED ISSUE #625: Feature request: support optional key `disabled`
- `disabled`: disable watch (default: False)

When set to `True`, do not watch the specified entry.

Jamstah:
This adds an `enabled` flag for all jobs that allows a user to disable individual jobs without having to comment out lots of lines of yaml.

Fixes #625

### DIFF
--- a/docs/source/jobs.rst
+++ b/docs/source/jobs.rst
@@ -176,6 +176,7 @@ Optional keys for all job types
 - ``compared_versions``: Number of versions to compare for similarity
 - ``kind`` (redundant): Either ``url``, ``shell`` or ``browser``.  Automatically derived from the unique key (``url``, ``command`` or ``navigate``) of the job type
 - ``user_visible_url``: Different URL to show in reports (e.g. when watched URL is a REST API URL, and you want to show a webpage)
+- ``enabled``: Can be set to false to disable an individual job (default is ``true``)
 
 
 Setting keys for all jobs at once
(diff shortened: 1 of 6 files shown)


docs/source/jobs.rst
```diff
@@ -176,6 +176,7 @@ Optional keys for all job types
 - ``compared_versions``: Number of versions to compare for similarity
 - ``kind`` (redundant): Either ``url``, ``shell`` or ``browser``.  Automatically derived from the unique key (``url``, ``command`` or ``navigate``) of the job type
 - ``user_visible_url``: Different URL to show in reports (e.g. when watched URL is a REST API URL, and you want to show a webpage)
+- ``enabled``: Can be set to false to disable an individual job (default is true)
```
thp, line 179:
```suggestion
- ``enabled``: Can be set to false to disable an individual job (default is ``true``)
```

lib/urlwatch/jobs.py
```diff
@@ -205,6 +205,9 @@ def location_is_url(self):
     def pretty_name(self):
         return self.name if self.name else self.get_location()
 
+    def is_enabled(self):
+        return self.enabled == None or self.enabled
```
thp, line 209:
```suggestion
        return self.enabled is None or self.enabled
```

thp (changes requested):
Very nice and complete PR, thanks!

Two commends + please update `CHANGELOG.md` with the additional feature.

Jamstah:
Done, thanks for the review.

Jamstah:
Should I also be regenerating the man pages, or does that get done at release time?

thp (approved):

thp:
Thanks!

Issue #507, opened 2020-05-29
How To Set A Reporter For A Job?


dinosaw:
Hello,

I am attempting to use the `slack` reporter. In the `--edit-config` file, I have set slack enabled to `true` and the `webhook_url` to the correct Slack URL. My question is now, how do I define it in the job? I don't see it described in the [readthedocs documentation](https://urlwatch.readthedocs.io/en/latest/reporters.html) anywhere.

I tried something similar to the following syntax:

```
name: "Test 1"
url: "https://google.com/"
filter:
  - html2text:
      method: pyhtml2text
report:
  - slack:
```

But that doesn't seem to work. I don't know where in the job's configuration I'm supposed to tell it to use the Slack reporter. Can someone provide me an example template of what theirs looks like?

Additionally, is it expected to receive the following error when running `--test-slack`:

```
Error while submitting message to Slack:{"message": "Cannot send an empty message", "code": 50006}
```

Thanks!

scottmac:
The reporters aren't job specific, it would send to all reporters.

Are you submitting to discord instead of slack? I think that error message is from discord.

thp:
Right now, reporters cannot be configured per-job. Closing as duplicate of #468.

thp:
Reopening and closing #468 instead.

teeann:
Is there any plan on resolving this? I think this use case is really essential for this tool.

thp:
To quote from the other issue:

> It can still be done by maintaining a separate config + job list and use `--config` and `--urls` with different invocations.

It might be essential for your use case (and you can work around it by specifying separate configs and url lists and running it for each set already now), but is *not* "essential for this tool". Different people use the tool differently.

For my use cases, I have a single reporter and multiple jobs, and that works just fine, and apparently this also applies to other users of this tool. I personally never had the need for setting different reporters for different jobs.

I'm sure that multiple people would welcome this change and have a use case for this, but this is not an "I'll scratch everyone's itch" project, it's an "I made this for myself, and if it's useful to anybody, that's great -- plus I'll accept pull requests if done well for things that other people might find useful and that fit the project's quality standards" project.

If you need this, feel free to implement it and send a pull request or if you don't have time or skills to implement it yourself, pay someone to do it for you.

In short, there's no plan on resolving this that I'm aware of at the moment. Patches gladly accepted.

teeann:
@thp Is it possible to add a FAQ section regarding this issue to the existing documentation? 
[...]

Issue #781, opened 2023-12-17
add support for specify multiple recipients per URL


monperrus:
Thanks for the great tool!

It would be super useful to specify different and possibly multiple recipients per URL.

for example:

```
url: http://example.com
to: me@foo.com,you@bar.com

url: http://easter.com
to: me@foo.com,friend@42.com
```

WDYT?

thp:
Right now, reporter-related configuration per job isn't possible. You have to rely on having multiple configurations and/or set up mailing lists or something. Because reports are grouped (so there's only one notification sent out if both are changed) it wouldn't even be possible without some additional logic to split reports in those cases (e.g. in your example, would me@foo.com get an e-mail with two changes, and you@bar.com and friend@42.com each get separate e-mails with their respective change?).

Possibly related to #507.

monperrus:
Thanks for the feedback.

> Because reports are grouped (so there's only one notification sent out if both are changed) it wouldn't even be possible without some additional logic to split reports in those cases

FYI, I have a working implementation to split reports in separate emails.

monperrus:
> Because reports are grouped

Trying to rebase before PRing, I discover the existence of new option "separate",
https://github.com/thp/urlwatch/blob/e342af925930114b4194f1bdb660dec6348f653a/lib/urlwatch/reporters.py#L130

Can this be used to send one email per report?

thp:
> > Because reports are grouped
> 
> Trying to rebase before PRing, I discover the existence of new option "separate",
> 
> https://github.com/thp/urlwatch/blob/e342af925930114b4194f1bdb660dec6348f653a/lib/urlwatch/reporters.py#L130
> 
> Can this be used to send one email per report?

This just sends multiple e-mails to the single recipient instead of a single mail with all update notifications combined. Yes, in this case it would be possible to have multiple recipients, but it means that having multiple recipients would enable `separate`, at least for that list of recipients, and then it can get tricky.

It's still tricky, as the way recipients are specified is specific to the reporter (e.g. e-mail reporter takes e-mail adresses, the Slack reporter takes a URL, the Telegram reporter takes something else, etc...).

monperrus:
I'm trying to use the `separate` option.

Neither top-level ` separate: true`.

nor under report works.

```yaml
report:
  email:
    enabled: true
    separate: true
``` 

any suggestion? Thanks!

monperrus:
Have a working version with adding the optional field `to` to a job: 
```yaml
---
kind: url
[...]

PR #789, opened 2024-01-25
Add tags to jobs and a command line arg to select them

Jamstah:
This allows the user to selectively run jobs based on a tag for use cases where you might want to run urlwatch on different schedules for different jobs and other use cases.

This is also part 1 of addressing #507 - next up would be allowing the report config to be an array (retaining existing behaviour for maps) and allowing users to assign tags to each reporter.

I am aware this will conflict with #785, and will rebase as required ;)


lib/urlwatch/config.py
```diff
@@ -68,6 +68,7 @@ def parse_args(self, cmdline_args):
         parser = argparse.ArgumentParser(description=urlwatch.__doc__,
                                          formatter_class=argparse.RawDescriptionHelpFormatter)
         parser.add_argument('joblist', metavar='JOB', type=int, nargs="*", help='index of job(s) to run, as numbered according to the --list command. If none specified, then all jobs will be run.')
+        parser.add_argument('--tags', metavar='TAG,TAG...', type=lambda x: x.split(","), help='comma separated tags of job(s) to run. If none specified, then all jobs will be run.')
```
thp, line 71: Shouldn't this be using [`nargs="+"`](https://docs.python.org/3/library/argparse.html#nargs), possibly with `action="extend"`?

thp (changes requested):
Needs rebasing.

Wonder if this tagging mechanism could be used for other things and have a bit broader scope? (e.g. reporters could also use tags, e.g for #507, #781 and #790)

lib/urlwatch/config.py
Jamstah: We already take a variable number of job indexes, so I thought nargs + would be confusing. How would you interpret `urlwatch --tags abc def 1 6`?

We could potentially say you can't use them both together, and that `--tags` changes from using job indexes to tags on the command line. Wdyt?

thp: Yeah, making `--tags` an `action="store_true"` and treating `joblist` as a list of tags if `args.tags` is `True` seems like a sensible idea. It then basically becomes a switch that changes the meaning of the positional arguments from job indexes to tags (the `joblist` help string might also need updating then).

Jamstah: I've done that, and refactored to clean it up a little. Squashed the push because it was a fairly heavy change to the approach.

lib/urlwatch/config.py
```diff
@@ -95,17 +96,27 @@ def parse_args(self, cmdline_args):
...
         group = parser.add_argument_group('miscellaneous')
         group.add_argument('--features', action='store_true', help='list supported jobs/filters/reporters')
         group.add_argument('--gc-cache', metavar='RETAIN_LIMIT', type=int, help='remove old cache entries, keeping the latest RETAIN_LIMIT (default: 1)',
                            nargs='?', const=1)
 
         args = parser.parse_args(cmdline_args)
 
-        for i, arg in enumerate(vars(args)):
+        if args.tags:
+            self.tag_set = frozenset(args.joblist)
```
thp, line 113: Should we warn/fail if `args.joblist` is empty? Not specifying jobs means "all jobs", but just specifying `--tags` without any jobs might warrant a failure?

lib/urlwatch/config.py
```diff
@@ -95,17 +96,27 @@ def parse_args(self, cmdline_args):
...
+        else:
+            try:
+                self.idx_set = frozenset([int(s) for s in args.joblist])
```
thp, line 116: I think `frozenset()` can accept a generator expression:

```suggestion
                self.idx_set = frozenset(int(s) for s in args.joblist)
```

lib/urlwatch/main.py
```diff
@@ -68,6 +68,20 @@ def __init__(self, urlwatch_config, config_storage, cache_storage, urls_storage)
...
+
+        if self.urlwatch_config.tags:
+            # If we're using tags, check the job matches the tags
+            return job.matches_tags(self.urlwatch_config.tag_set)
+        else:
+            # If we're using indexes, no indexes means all jobs
+            if self.urlwatch_config.idx_set:
+                return idx in self.urlwatch_config.idx_set
+            else:
+                return True
```
thp, line 83: When the last statement in an `if` block is a `return`, I like to save some indentation using:

```suggestion
        if self.urlwatch_config.tags:
            # If we're using tags, check the job matches the tags
            return job.matches_tags(self.urlwatch_config.tag_set)

        # If we're using indexes, no indexes means all jobs
        if self.urlwatch_config.idx_set:
            return idx in self.urlwatch_config.idx_set
            
        return True
```

lib/urlwatch/tests/test_handler.py
```diff
@@ -99,8 +99,8 @@ def py_files():
 
 
 class ConfigForTest(CommandConfig):
-    def __init__(self, config, urls, cache, hooks, verbose):
-        super().__init__([], 'urlwatch', os.path.dirname(__file__), root, config, urls, hooks, cache, verbose)
+    def __init__(self, config, urls, cache, hooks, verbose, args=[]):
```
thp, line 102: `args=[]` is a [mutable default argument](https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments).

Something immutable (e.g. `args=frozenset()` or the empty tuple `args=()`) would be totally acceptable, though.

thp (changes requested):
Added some comments. And as you mentioned, needs rebasing.

lib/urlwatch/tests/test_handler.py
Jamstah: Every day is a school day, thanks

lib/urlwatch/config.py
Jamstah: As docced (in this pr) we say specifying nothing means run all jobs, so I'm tempted to go with that behaviour.

Although from a Unix clean interface standpoint, I kind of prefer that it just doesn't run anything and "no tags" isn't a special case. The less special cases the less surprises.

I think the "run all" behaviour is easier to explain (document).

Jamstah: I added a test for the behaviour in the base commit, and a new commit on top to switch to running all for the empty list.

If you still think the behaviour should be to error instead, just shout and I'll switch it to that.

docs/source/manpage.rst
```diff
@@ -23,13 +23,17 @@ This manpage describes the CLI tool.
 
 positional arguments:
    JOB
-          index of job(s) to run, as numbered according to the --list command.
+          indexes or tags of job(s) to run, depending on --tags.
+          If using indexes, they are as numbered according to the --list command.
```
thp, line 27: Something along the lines of:

```suggestion
          index(es) or tag(s) of job(s) to run.
          If --tags is set, JOB will be interpreted as tags,
          if not, as index(es) numbered according to the --list command.
```

lib/urlwatch/config.py
thp: > If you still think the behaviour should be to error instead, just shout and I'll switch it to that.

Yes, I think failing loudly is good here, otherwise people will have things like `urlwatch --tags $SOME_VARIABLE` and if `$SOME_VARIABLE` is empty (e.g. mistyped or empty for other reasons), then this will not do what the user expects.

What would be the use case of doing nothing at all when running `urlwatch --tags`?

lib/urlwatch/jobs.py
```diff
@@ -196,7 +197,10 @@ def ignore_error(self, exception):
 
 class Job(JobBase):
     __required__ = ()
-    __optional__ = ('name', 'filter', 'max_tries', 'diff_tool', 'compared_versions', 'diff_filter', 'enabled', 'treat_new_as_changed', 'user_visible_url')
+    __optional__ = ('name', 'tags', 'filter', 'max_tries', 'diff_tool', 'compared_versions', 'diff_filter', 'enabled', 'treat_new_as_changed', 'user_visible_url')
```
thp, line 200: Nitpicky, but:

```suggestion
    __optional__ = ('name', 'filter', 'max_tries', 'diff_tool', 'compared_versions', 'diff_filter', 'enabled', 'treat_new_as_changed', 'user_visible_url', 'tags')
```

lib/urlwatch/jobs.py
```diff
@@ -196,7 +197,10 @@ def ignore_error(self, exception):
...
+
+    def matches_tags(self, tags: Set[str]):
+        return self.tags & tags
```
thp, line 203: Is it `matches_tags` or `matching_tags`? Could have a type annotation for the return type?

lib/urlwatch/jobs.py
```diff
@@ -208,6 +212,17 @@ def pretty_name(self):
...
+    @property
+    def tags(self) -> Optional[FrozenSet[str]]:
+        return self._tags
+
+    @tags.setter
+    def tags(self, value: Optional[Iterable[str]]):
+        if value is None:
+            self._tags = None
+        else:
+            self._tags = frozenset(value)
```
thp, line 224: Personal preference:

```suggestion
        if value is None:
            self._tags = None
        
        self._tags = frozenset(value)
```

lib/urlwatch/worker.py
```diff
@@ -51,11 +51,13 @@ def run_parallel(func, items):
...
 
 def run_jobs(urlwatcher):
-    if not all(1 <= idx <= len(urlwatcher.jobs) for idx in urlwatcher.urlwatch_config.joblist):
+    if not urlwatcher.urlwatch_config.tags and not all(1 <= idx <= len(urlwatcher.jobs) for idx in urlwatcher.urlwatch_config.idx_set):
         raise ValueError(f'All job indices must be between 1 and {len(urlwatcher.jobs)}: {urlwatcher.urlwatch_config.joblist}')
     cache_storage = urlwatcher.cache_storage
     jobs = [job.with_defaults(urlwatcher.config_storage.config)
-            for (idx, job) in enumerate(urlwatcher.jobs) if job.is_enabled() and ((idx + 1) in urlwatcher.urlwatch_config.joblist or (not urlwatcher.urlwatch_config.joblist))]
+            for (idx, job) in enumerate(urlwatcher.jobs)
+            if urlwatcher.should_run(idx, job)
```
thp, line 59: Shouldn't this have `idx + 1`?

lib/urlwatch/main.py
```diff
@@ -68,6 +68,21 @@ def __init__(self, urlwatch_config, config_storage, cache_storage, urls_storage)
         if hasattr(self.urlwatch_config, 'migrate_urls'):
             self.urlwatch_config.migrate_cache(self)
 
+    def should_run(self, idx, job):
```
thp, line 71: This is nice, as the expression in the `worker.py` list comprehension was already getting kind of unwieldy, thanks :)

thp (changes requested):
Thanks for working on this! See comments.

lib/urlwatch/jobs.py
Jamstah: There's no return, so that would fall through and attempt to make a frozenset with None in it.

lib/urlwatch/worker.py
Jamstah: Great catch!

lib/urlwatch/jobs.py
Jamstah: I've moved it, but why at the end? Its not alphanum sorted, and I thought putting it with `name` made sense as they're both identifying metadata.

lib/urlwatch/config.py
Jamstah: The only unix style reason to not error to I can think of is:
```
./tags-to-run.sh | xargs urlwatch --tags
```

And even that can be solved with `xargs --no-run-if-empty`

Anyway, I'm splitting hairs, erroring on an empty tag list is perfectly acceptable behaviour. Have done that.

docs/source/manpage.rst
Jamstah: Done something along those lines :)

thp (approved):
Thanks!

PR #794, opened 2024-02-05
Adds optional reply_to email reporter config option

trevorshannon:
It can sometimes be useful to have a reply-to address different from the from address, especially when sending automated emails such as those generated by urlwatch.  If email reports are being sent to a distribution list, a reply-to address of that list makes more sense than the (likely unmonitored) bot address.

This PR adds the `reply_to` config option for the email reporter.  Note that the `Reply-To` header will now always be added to outgoing emails, but if the `reply_to` option is missing, that header will simply be populated with the same value as `From`


Jamstah:
You should also update the CHANGELOG.md.

trevorshannon:
oops, thanks I always forget that.

thp (approved):
Very nice! And I agree that specifying a `reply_to` value should cause the header to be added, even if it's the same as the `from` value, good point.

PR #797, opened 2024-02-22
add support for sending to multiple recipients via sendmail

monperrus:
if the `to` config contains multiple emails such as `x@foo.com,y@bar.com`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,7 +18,7 @@ The format mostly follows [Keep a Changelog](http://keepachangelog.com/en/1.0.0/
 
 ### Fixed
 
-- Fix multiple recipients for sendmail reporter (#797)
+- `email` reporter: Allow multiple recipients for `sendmail` method (#797, by monperrus)
 - Fix documentation for watching Github tags and releases, again (#723)
 - Fix `--test-reporter` command-line option so `separate` configuration option is no longer ignored when sending test notifications (#772, by marunjar)
 - Fix line height and dark mode regression (#774 reported by kongomongo, PRs #777 and #778 by trevorshannon)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ The format mostly follows [Keep a Changelog](http://keepachangelog.com/en/1.0.0/
 
 ### Fixed
 
+- Fix multiple recipients for sendmail reporter (#797)
 - Fix documentation for watching Github tags and releases, again (#723)
 - Fix `--test-reporter` command-line option so `separate` configuration option is no longer ignored when sending test notifications (#772, by marunjar)
 - Fix line height and dark mode regression (#774 reported by kongomongo, PRs #777 and #778 by trevorshannon)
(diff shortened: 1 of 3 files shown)


Jamstah:
Does this also work for the smtp method? That field should probably work the same for both mechanisms.

This PR is missing documentation, and an entry in the changelog.

monperrus:
> Does this also work for the smtp method?

Yes, it already works for SMTP

> This PR is missing documentation and an entry in the changelog.

Done.

Thanks!

docs/source/reporters.rst
```diff
@@ -69,6 +69,7 @@ At the moment, the following reporters are built-in:
 - **prowl**: Send a detailed notification via prowlapp.com
 - **pushbullet**: Send summary via pushbullet.com
 - **pushover**: Send summary via pushover.net
+- **sendmail**: Pipe a message to the system sendmail command
```
Jamstah, line 72: There isn't a reporter called `sendmail`, its a method of the `email` reporter.

I think we need a new section like [this one](https://urlwatch.readthedocs.io/en/latest/reporters.html#e-mail-via-gmail-smtp), but titled "E-Mail via sendmail" or something.

Jamstah (changes requested):

docs/source/reporters.rst
monperrus: @Jamstah done!

docs/source/reporters.rst
```diff
@@ -290,6 +290,20 @@ public Matrix room, as the messages quickly become noisy:
...
+---------------------
+
+You can send email via the system's sendmail command provided by the MTA. You need to set "method: sendmail" in the config file.
+
+.. code:: yaml
+
+    report:
+      email:
+        enabled: true
+        from: 'postmaster@example.com'
```
Jamstah, line 303: Should probably have a `to:` in this example as well.

monperrus: done

CHANGELOG.md
```diff
@@ -18,6 +18,7 @@ The format mostly follows [Keep a Changelog](http://keepachangelog.com/en/1.0.0/
 
 ### Fixed
 
+- Fix multiple recipients for sendmail reporter (#797)
```
thp, line 21:
```suggestion
- `email` reporter: Allow multiple recipients for `sendmail` method (#797, by monperrus)
```

docs/source/reporters.rst
```diff
@@ -290,6 +290,21 @@ public Matrix room, as the messages quickly become noisy:
      minimal: true
      enabled: true
 
+E-Mail via sendmail
+---------------------
+
+You can send email via the system's sendmail command provided by the MTA. You need to set "method: sendmail" in the config file.
```
thp, line 296: Formatting nitpicks:
 
```suggestion
You can send email via the system's ``sendmail`` command provided by the MTA. You need to set ``method: sendmail`` in the config file:
```

thp (changes requested):
Thanks, two small comments.

monperrus:
Thanks for the review, applied the changes.

thp (approved):
Thanks!

monperrus:
thank you!

PR #801, opened 2024-02-26
Fix IndexError after failed edit

jwilk:
Fixes:

    IndexError: string index out of range

if the user just pressed enter after the "Do you want to retry the same edit?" question.


Jamstah:
Can you please add to the CHANGELOG file as well?

thp (changes requested):
The thing is, what do we expect to happen when the user just presses Enter?

If it should default to the "yes" answer (sounds reasonable), we should also change the prompt to:

    Do you want to retry the same edit? (Y/n)

And as @Jamstah mentioned, a changelog entry might be warranted (even though it's quite minor that it shouldn't affect users upgrading from an older version negatively, and it's an interactive thing).

Jamstah:
Thanks for the heads up. I've built on your commit in #812 to address the review comments.

PR #805, opened 2024-03-09
Add re.findall to pick out re matches

### LINKED ISSUE #804: Feature request: Extension of regex filtering to extract data
I couldn't find a way, how to use `re.sub` to extract data as a filter in urlwatch. Unless I haven't overseen anything, and there is a way to do it, I would like to open a feature request.

According to #401 I would like to request an extensions of the regex filtering. Instead of replacing matched strings, it would be useful to have a positive regex filter, to extract specific data. An example of this filtering is even in the [documentation](https://urlwatch.readthedocs.io/en/latest/filters.html#using-a-shell-script-as-a-filter). However there it is implemented using shellpipes and grep, but for me it turned out to be a bit unstable.

```
url: https://example.net/shellpipe-grep.txt
filter:
  - shellpipe: "grep -i -o 'price: <span>.*</span>'"
```

IMO better approach would be to have a filter like `re.findall` which goes along with `re.sub`:

```
url: https://example.net/pricelist.html
filter:
  - re.findall: 'price: <span>(.*)</span>'
```

Jamstah:
Actually using re.finditer so we can apply a repl to the result. This allows users to pick out matches and reformat them in one step.

Fixes #804

### DIFF
--- a/docs/source/filters.rst
+++ b/docs/source/filters.rst
@@ -77,6 +77,7 @@ At the moment, the following filters are built-in:
 - **ical2text**: Convert `iCalendar`_ to plaintext
 - **ocr**: Convert text in images to plaintext using Tesseract OCR
 - **re.sub**: Replace text with regular expressions using Python's re.sub
+- **re.findall**: Find all non-overlapping matches using Python's re.findall
 - **reverse**: Reverse input items
 - **sha1sum**: Calculate the SHA-1 checksum of the content
 - **shellpipe**: Filter using a shell command
@@ -485,12 +486,13 @@ Alternatively, ``jq`` can be used for filtering:
    filter:
      - jq: '.[0].name'
 
-Remove or replace text using regular expressions
-------------------------------------------------
+Find, remove or replace text using regular expressions
+------------------------------------------------------
 
-Just like Python’s ``re.sub`` function, there’s the possibility to apply
-a regular expression and either remove of replace the matched text. The
-following example applies the filter 3 times:
+You can use ``re.sub`` and ``re.findall`` to apply regular expressions.
+
+``re.sub`` can be used to remove or replace all non-overlapping instances
+of matched text. The following example applies the filter 3 times:
 
 1. Just specifying a string as the value will replace the matches with
    the empty string.
@@ -499,11 +501,7 @@ following example applies the filter 3 times:
 3. You can use groups (``()``) and back-reference them with ``\1``
    (etc..) to put groups into the replacement string.
 
-All features are described in Python’s
-`re.sub <https://docs.python.org/3/library/re.html#re.sub>`__
-documentation (the ``pattern`` and ``repl`` values are passed to this
-function as-is, with the value of ``repl`` defaulting to the empty
-string).
+``repl`` defaults to the empty string, which will remove matched strings.
 
 .. code:: yaml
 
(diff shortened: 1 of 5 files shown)


Jamstah:
Draft because it needs documentation.

docs/source/filters.rst
```diff
@@ -517,15 +515,35 @@ string).
...
+2. You can use groups (``()``) and back-reference them with ``\1``
+   (etc..) to put groups into the replacement string.
+
+By default, the full match will be included in the output.
+
+.. code:: yaml
+
+   url: https://example.com/regex-findall.html
+   filter:
+       - re.findall: '<span class="price">.*</span>'
```
thp, line 533: This is a bad example. The CSS or XPath selector is more suited for this. From the [obligatory Stack Overflow](https://stackoverflow.com/a/1732454/1047040):

> You can't parse [X]HTML with regex. Because HTML can't be parsed by regex. Regex is not a tool that can be used to correctly parse HTML. As I have answered in HTML-and-regex questions here so many times before, the use of regex will not allow you to consume HTML.

thp (changes requested):
In general, I don't mind a `re.findall()`, but people might be tempted to use it to "parse" HTML (which is not possible using regex) when the CSS or XPath filters are more suited for the job.

If we make it clear in the docs when it isn't and when it is the right tool for the job, I'm happy to merge this :)

Jamstah:
Yeah, I was thinking that but was too lazy to think of a good example :)

I've borrowed something from another test, and added some doc.

lib/urlwatch/filters.py
```diff
@@ -848,6 +848,26 @@ def filter(self, data, subfilter):
...
+    }
+
+    __default_subfilter__ = 'pattern'
+
+    def filter(self, data, subfilter):
+        if 'pattern' not in subfilter:
+            raise ValueError('{} needs a pattern'.format(self.__kind__))
+
+        # Default: Replace with full match if no "repl" value is set
+        return "\n".join([match.expand(subfilter.get('repl', '\\g<0>')) for match in re.finditer(subfilter['pattern'], data)])
```
thp, line 868: Generator expression should be fine?

```suggestion
        return "\n".join(match.expand(subfilter.get('repl', '\\g<0>')) for match in re.finditer(subfilter['pattern'], data))
```

thp (changes requested):
See comments.

lib/urlwatch/filters.py
```diff
@@ -848,6 +848,26 @@ def filter(self, data, subfilter):
...
 
 
+class RegexFindall(FilterBase):
+    """Pick out regular expressions using Python's re.findall"""
+
+    __kind__ = 're.findall'
+
+    __supported_subfilters__ = {
+        'pattern': 'Regular expression to search for (required)',
+        'repl': 'Replacement string (default: empty string)',
```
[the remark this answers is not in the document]
Jamstah: Thanks, fixed.

docs/source/filters.rst
```diff
@@ -485,12 +486,13 @@ Alternatively, ``jq`` can be used for filtering:
...
-------------------------------------------------
+Find, remove or replace text using regular expressions
+------------------------------------------------------
 
-Just like Python’s ``re.sub`` function, there’s the possibility to apply
-a regular expression and either remove of replace the matched text. The
-following example applies the filter 3 times:
+You can use ``re.sub`` and ``re.findall`` to apply regular expressions.
+
+``re.sub`` can be used to remove of replace all non-overlapping instances
```
thp, line 494:
```suggestion
``re.sub`` can be used to remove or replace all non-overlapping instances
```

docs/source/filters.rst
```diff
@@ -517,15 +515,40 @@ string).
...
+2. It breaks the numbers in the date out into separate lines.
+
+If ``repl`` is not specified, the full match will be included in the output.
+
+.. code:: yaml
+
+   url: https://example.com/regex-findall.html
+   filter:
+       - re.findall:
+           pattern: 'The next draw is on (\d\d\d\d-\d\d-\d\d).'
```
thp, line 533: Maybe:

```suggestion
           pattern: 'The next draw is on (\d{4}-\d{2}-\d{2}).'
```

docs/source/filters.rst
```diff
@@ -517,15 +515,40 @@ string).
...
+           repl: '\1'
+       - re.findall: '[0-9]+'
```
thp, line 535: Maybe:
```suggestion
       - re.findall: '\d+'
```

docs/source/filters.rst
```diff
@@ -517,15 +515,40 @@ string).
...
+
+Note: When using HTML or XML, it is usually better to use CSS selectors or
+XPATH expressions. HTML and XML cannot be parsed properly using regular
+expressions. If the CSS selector or XPATH cannot provide the targeted
```
thp, line 539: Could add a link to https://stackoverflow.com/a/1732454/1047040 on "cannot be parsed" for extra entertainment.

thp (changes requested):
Few comments, rest looks good. Thanks!

docs/source/filters.rst
Jamstah: It is a classic :)

thp (approved):
Thanks!

PR #810, opened 2024-03-22
Add a wait_for option to browser jobs

### LINKED ISSUE #763: Wait for selector?
As #548, some websites don't load properly (one serious example would be jp.mercari.com/) with just a `wait_until`. I think it'd be better to enable users to use `wait_for` to wait for a selector, or enables users to inject/eval a custom JavaScript script within the filter chain so it allows users much more capability more than the waiting issue. Thanks.

Jamstah:
This waits for the default timeout of 30 seconds for the selector to be satisfied. The selector is based on
https://playwright.dev/python/docs/locators#locate-by-css-or-xpath.

Closes #763


thp (approved):
Thanks!

PR #812, opened 2024-04-04
Fix IndexError after failed edit

Jamstah:
Follow on from #801


thp (approved):
Thanks for adopting and adapting this PR!

PR #822, opened 2024-05-29
Allow reporters to be specified multiple times

### LINKED ISSUE #790: FEATURE: Support multiple reporters with different options
Interested in thoughts of this style of config

```yaml
report:
  text: # Still have these for inheritance
    details: true
    footer: true
    line_length: 75
    minimal: false
  html:
    diff: unified
  reporters:
  - type: email
    enabled: true
    from: your.username@gmail.com
    to: person1@domain.net
    method: sendmail
    tags: tag1
  - type: email
    enabled: true
    from: your.username@gmail.com
    to: person2@domain.net
    method: sendmail
    tags: tag2
    html: true
  - type: stdout
    color: true
    enabled: true
  telegram: # Still support dict format
    bot_token: ''
    chat_id: ''
    enabled: false
```

Jamstah:
This allows different jobs to be selected for reporting in different ways, for example, allowing different changes to be emailed to different addresses.

Closes #790


lib/urlwatch/reporters.py
```diff
@@ -96,8 +102,10 @@ def get_signature(self):
                                                       copyright=urlwatch.__copyright__),
             'Website: {url}'.format(url=urlwatch.__url__),
             'Support urlwatch development: https://github.com/sponsors/thp',
-            'watched {count} URLs in {duration} seconds'.format(count=len(self.job_states),
-                                                                duration=self.duration.seconds),
+            'reported {count} of {total} watched URLs in {duration} seconds'.format(
+                count=len(self.job_states),
+                total=self.job_count_total,
+                duration=self.duration.seconds),
```
thp, line 108: The "reported in X seconds" is kind of wrong, as most of the time will be spent watching (hopefully).

Maybe just leave this as-is, as I'm not sure how useful this is (apart from debugging, which is already in the logs?).

thp (changes requested):
See comments.

lib/urlwatch/reporters.py
Jamstah: Have removed the count, it was hard to describe what the count meant. Instead, we can just say what the process did.

I think that's clearer, or, we could just remove the sentence altogether.

lib/urlwatch/command.py
```diff
@@ -380,7 +380,10 @@ def set_error(job_state, message):
                                    'Same Old, Same Old\n'))
         report.error(set_error(build_job('Error Reporting', 'http://example.com/error', '', ''), 'Oh Noes!'))
 
-        report.finish_one(name)
+        reported = report.finish_one(name)
+
+        if not reported:
+            raise ValueError(f'Reporter not enabled: {name}')
```
thp, line 386:
```suggestion
        if not report.finish_one(name):
            raise ValueError(f'Reporter not enabled: {name}')
```

..but is that changing behaviour from before?

lib/urlwatch/main.py
```diff
@@ -109,5 +109,13 @@ def run_jobs(self):
         run_jobs(self)
 
     def close(self):
-        self.report.finish()
+        reported = self.report.finish()
+
+        if not reported:
+            logger.warning('No reporters enabled.')
```
thp, line 115: I like to do it like this mostly so that the "reported" local variable doesn't leak into the scope (this way, we can be sure that it isn't used below, since we don't give the expression result a name to begin with):

```suggestion
        if not self.report.finish():
            logger.warning('No reporters enabled.')
```

lib/urlwatch/reporters.py
```diff
@@ -123,35 +133,36 @@ def reporter_documentation(cls):
...
-    def submit_all(cls, report, job_states, duration):
-        any_enabled = False
-        for name, subclass in cls.__subclasses__.items():
-            cfg = report.config['report'].get(name, {})
+        for cfg in cfgs:
             if cfg.get('enabled', False):
                 any_enabled = True
                 logger.info('Submitting with %s (%r)', name, subclass)
                 base_config = subclass.get_base_config(report)
+                matching_job_states = filter_by_tags(job_states, cfg.get("tags", []))
```
thp, line 147: I wonder if we can get away with this, and then not having to check for the empty tags in `filter_by_tags()`:

```suggestion
                if tags := cfg.get("tags"):
                    job_states = filter_by_tags(job_states, tags)
```

lib/urlwatch/reporters.py
```diff
@@ -123,35 +133,36 @@ def reporter_documentation(cls):
...
                 if base_config.get('separate', False):
-                    for job_state in job_states:
-                        subclass(report, cfg, [job_state], duration).submit()
+                    for job_state in matching_job_states:
+                        subclass(report, cfg, [job_state], len(job_states), duration).submit()
+                        job_state.reported_count = job_state.reported_count + 1
                 else:
-                    subclass(report, cfg, job_states, duration).submit()
+                    subclass(report, cfg, matching_job_states, len(job_states), duration).submit()
```
thp, line 153: I haven't checked what's the right way, but curiously it's always `len(job_states)`, not `len(matching_job_states)`, is that correct/intended or a bug?

lib/urlwatch/reporters.py
```diff
@@ -123,35 +133,36 @@ def reporter_documentation(cls):
...
+                    for job_state in matching_job_states:
+                        job_state.reported_count = job_state.reported_count + 1
```
thp, line 155:
```suggestion
                        job_state.reported_count += 1
```

lib/urlwatch/reporters.py
```diff
@@ -123,35 +133,36 @@ def reporter_documentation(cls):
...
+
+        return any_enabled
+
+    @classmethod
+    def submit_all(cls, report, job_states, duration):
+        any_enabled = False
+        for name in cls.__subclasses__.keys():
```
thp, line 162: This might also work:

```suggestion
        for name in cls.__subclasses__:
```

lib/urlwatch/reporters.py
```diff
@@ -123,35 +133,36 @@ def reporter_documentation(cls):
...
+            any_enabled = any_enabled | ReporterBase.submit_one(name, report, job_states, duration)
```
thp, line 163:
```suggestion
            any_enabled |= ReporterBase.submit_one(name, report, job_states, duration)
```

lib/urlwatch/reporters.py
```diff
@@ -96,8 +104,10 @@ def get_signature(self):
                                                       copyright=urlwatch.__copyright__),
             'Website: {url}'.format(url=urlwatch.__url__),
             'Support urlwatch development: https://github.com/sponsors/thp',
-            'watched {count} URLs in {duration} seconds'.format(count=len(self.job_states),
-                                                                duration=self.duration.seconds),
+            'watched {total} URLs in {duration} seconds'.format(
```
thp, line 107: Does it make sense to show here that something is filtered down? e.g. `watched 5 or 9 URLs in 12.34 seconds`?

thp (changes requested):
See comments.

lib/urlwatch/command.py
Jamstah: You mean is raising an error if there are no reporters enabled changing the behaviour?

Yep, but I would argue that anyone running it without reporting anything is probably doing something wrong...

lib/urlwatch/jobs.py
```diff
@@ -199,10 +199,8 @@ class Job(JobBase):
     __required__ = ()
     __optional__ = ('name', 'filter', 'max_tries', 'diff_tool', 'compared_versions', 'diff_filter', 'enabled', 'treat_new_as_changed', 'user_visible_url', 'tags')
 
-    def matching_tags(self, tags: Set[str]) -> Set[str]:
-        if self.tags is None:
-            return False
-        return self.tags & tags
+    def matching_tags(self, tags: Iterable[str]) -> Set[str]:
+        return self.tags.intersection(tags)
```
thp, line 203: Looking at the old code, does this need to be:


```suggestion
    def matching_tags(self, tags: Iterable[str]) -> Set[str]:
        return self.tags is not None and self.tags.intersection(tags)
```

lib/urlwatch/main.py
```diff
@@ -110,4 +110,11 @@ def run_jobs(self):
         self.report.finish()
 
     def close(self):
+        if not self.report.reporters_enabled:
+            logger.warning('No reporters enabled.')
+
+        for job_state in self.report.job_states:
+            if not job_state.reported_count:
+                logger.warning(f'Job {job_state.job.pretty_name()} was not reported on')
```
thp, line 118: When will this warning be printed? I guess it won't get reported for unchanged jobs (since the job state isn't in `self.report.job_states`)?

lib/urlwatch/reporters.py
```diff
@@ -80,13 +81,20 @@
...
+    if tags:
+        return [job_state for job_state in job_states if job_state.job.matching_tags(tags)]
+    return job_states
+
+
 class ReporterBase(object, metaclass=TrackSubClasses):
     __subclasses__ = {}
 
-    def __init__(self, report, config, job_states, duration):
+    def __init__(self, report, config, job_states, job_count_total, duration):
```
thp: Add new parameters to the end of the parameter list; otherwise if someone accidentally mixes two versions of files (caller and callee), the error message will be misleading (old caller passes duration into `job_count_total` of new callee and Python reports that "`duration` is missing", or new caller passes total job count into `duration`, and Python reports "extra parameter `duration`" -- compared to "extra parameter `job_count_total`" vs "missing parameter `job_count_total`".

This is especially important, as theoretically one could implement other reporters out-of-tree, and those would break with this change.

lib/urlwatch/reporters.py
thp: Can probably be an empty tuple to guarantee it's immutable:

```python3
matching_job_states = filter_by_tags(job_states, cfg.get("tags", ()))
```

or given how `filter_by_tags()` is currently implemented, this should also work:

```python3
matching_job_states = filter_by_tags(job_states, cfg.get("tags", None))
```

And that's probably basically:

```python3
matching_job_states = filter_by_tags(job_states, cfg.get("tags"))
```

thp (changes requested):
Getting there. Some minor comments.

lib/urlwatch/main.py
Jamstah: It will get printed if a job state has changed, but that state change isn't reported on.

This is to make sure that if a change isn't reported on it doesn't get missed silently.

lib/urlwatch/jobs.py
Jamstah: All calls to matching_tags have already done the None check - we could do it again or we could remove the check from callers and return a falsey value here. Do you have a preference?

thp: In that case, it should be fine.

lib/urlwatch/main.py
thp: Great, let's do it like in the PR then :+1:

Jamstah: Scratch that, unchanged jobs are in job_states.

This will get printed if a URL was tested but matched no reporters. For example, if:
- My only reporter is stdout
- The stdout reporter has `tags: ["stdout"]`
- I have a job without the `stdout` tag

Then I'll see the warning.

lib/urlwatch/reporters.py
Jamstah: Not sure I should have resolved this - new change here that I hope is clearer.

lib/urlwatch/reporters.py
```diff
@@ -96,8 +104,11 @@ def get_signature(self):
                                                       copyright=urlwatch.__copyright__),
             'Website: {url}'.format(url=urlwatch.__url__),
             'Support urlwatch development: https://github.com/sponsors/thp',
-            'watched {count} URLs in {duration} seconds'.format(count=len(self.job_states),
-                                                                duration=self.duration.seconds),
+            '{total} URLs tested in {duration} seconds'.format(
+                total=self.job_count_total,
+                duration=self.duration.seconds),
+            'This report covers {count} URLs'.format(
+                count=len(self.job_states)),
```
thp: Minor nitpick:

```suggestion
            'watched {total} URLs in {duration} seconds'.format(
                total=self.job_count_total,
                duration=self.duration.seconds),
            'this report covers {count} URLs'.format(
                count=len(self.job_states)),
```

docs/source/reporters.rst
```diff
@@ -51,6 +51,34 @@ If the notification does not work, check your configuration and/or add
 the ``--verbose`` command-line option to show detailed debug logs.
 
 
+Common options
```
thp, line 54: Let's rewrite that into a section for the existing feature ("Common options" - nice find, seems we had that `enabled` config feature [since like forever](https://github.com/thp/urlwatch/commit/68329e08010e2c019decc472c2797da4faba8617) but it was never documented) and a section for the "meat" of this pull request, the multiple reporter feature:

```
Common options
--------------

The following configuration options can be applied to any reporter type:

* ``enabled``: Disable or enable a reporter


Configuring multiple reporters of the same type
-----------------------------------------------

You can use a list of configurations under a reporter type to report
different jobs with different configurations. You can select the jobs
for each reporter by using tags.

For example:

[...]
```

thp (changes requested):
Only minor rewording requests, the rest looks good now. Thanks for working on this!

docs/source/reporters.rst
Jamstah: Added tags as the second common option and put in a better example for configuring multiple reporters.

thp (approved):
Looks clean now.

Thanks for unifying `submit_all()` and `submit_one()`, that's a nice cleanup I just now saw.

PR #824, opened 2024-06-24
If tags is a string, use it as a single tag

Jamstah:
I accidentally forgot to make it an array in my yaml and nothing complained, but it wasn't working.


thp:
Thanks!

PR #831, opened 2024-11-10
Add new option to run just new jobs or jobs without history

nille02:
I have the issue that i have some url.yaml files with hundreds to thousands of jobs. A full run can take hours and i do that in general just once a month or even less. 
Before that option i checked for the new ids and run them manual. but it was quite the chore and i looked how i can avoid that. 

My only issue is that get_history_data() is quite slow but i guess sqllite is the reason?

### DIFF
--- a/lib/urlwatch/command.py
+++ b/lib/urlwatch/command.py
@@ -142,6 +142,19 @@ class UrlwatchCommand:
         # (ignore_cached) and we do not want to store the newly-retrieved data yet (filter testing)
         return 0
 
+    def prepare_jobs(self):
+        new_jobs = []
+        for idx, job in enumerate(self.urlwatcher.jobs):
+            has_history = self.urlwatcher.cache_storage.has_history_data(job.get_guid())
+            if not has_history:
+                logger.info('Add Job: %s', job.pretty_name())
+                new_jobs.append(idx + 1)
+        if not new_jobs:
+            return 0
+        self.urlwatch_config.idx_set = frozenset(new_jobs)
+        self.urlwatcher.run_jobs()
+        self.urlwatcher.close()
+
     def _resolve_job_history(self, id, max_entries=10):
         job = self._get_job(id)
 
@@ -274,6 +287,8 @@ class UrlwatchCommand:
             sys.exit(self.test_filter(self.urlwatch_config.test_filter))
         if self.urlwatch_config.test_diff_filter:
             sys.exit(self.test_diff_filter(self.urlwatch_config.test_diff_filter))
+        if self.urlwatch_config.prepare_jobs:
+            sys.exit(self.prepare_jobs())
         if self.urlwatch_config.dump_history:
             sys.exit(self.dump_history(self.urlwatch_config.dump_history))
         if self.urlwatch_config.list:
--- a/lib/urlwatch/config.py
+++ b/lib/urlwatch/config.py
@@ -93,6 +93,7 @@ class CommandConfig(BaseConfig):
         group.add_argument('--delete', metavar='JOB', help='delete job by location or index')
         group.add_argument('--enable', metavar='JOB', help='enable job by location or index')
         group.add_argument('--disable', metavar='JOB', help='disable job by location or index')
+        group.add_argument('--prepare-jobs', action='store_true', help='run jobs without history')
         group.add_argument('--change-location', metavar=('JOB', 'NEW_LOCATION'), nargs=2, help='change the location of an existing job by location or index')
         group.add_argument('--test-filter', metavar='JOB', help='test filter output of job by location or index')
         group.add_argument('--test-diff-filter', metavar='JOB',
--- a/lib/urlwatch/storage.py
+++ b/lib/urlwatch/storage.py
@@ -567,6 +567,8 @@ class CacheMiniDBStorage(CacheStorage):
         self.db = minidb.Store(self.filename, debug=True, vacuum_on_close=False)
         self.db.register(CacheEntry)
 
+        self._cached_has_history_data_set = None
+
     def close(self):
         self.db.close()
         self.db = None
@@ -599,6 +601,15 @@ class CacheMiniDBStorage(CacheStorage):
                 if len(history) >= count:
                     break
         return history
+
+    def has_history_data(self, guid):
+        if not self._cached_has_history_data_set:
+            self._cached_has_history_data_set = frozenset(guid[0] for guid in
+                                                           list(CacheEntry.query(self.db, CacheEntry.c.guid,
+                                                                                 where=((CacheEntry.c.tries == 0)
+                                                                                        | (CacheEntry.c.tries == None)))  # noqa:E711
+                                                                ))
+        return guid in self._cached_has_history_data_set
 
     def save(self, job, guid, data, timestamp, tries, etag=None):
         self.db.save(CacheEntry(guid=guid, timestamp=timestamp, data=data, tries=tries, etag=etag))
@@ -687,6 +698,9 @@ class CacheRedisStorage(CacheStorage):
                     if len(history) >= count:
                         break
         return history
+
+    def has_history_data(self, guid):
+        return bool(self.get_history_data(guid))
 
     def save(self, job, guid, data, timestamp, tries, etag=None):
         r = {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,12 @@ All notable changes to this project will be documented in this file.
 
 The format mostly follows [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 
+## UNRELEASED
+
+### Added
+
+- New command-line option `--prepare-jobs` to initialize new jobs or jobs without history (#831 by nille02)
+
 ## [2.29] -- 2024-10-28
 
 ### Added


lib/urlwatch/command.py
```diff
@@ -142,6 +142,21 @@ def test_filter(self, id):
         # (ignore_cached) and we do not want to store the newly-retrieved data yet (filter testing)
         return 0
 
+    def run_new_jobs(self):
+        new_jobs = list()
+        for idx, job in enumerate(self.urlwatcher.jobs):
+            history_data = self.urlwatcher.cache_storage.get_history_data(job.get_guid(), 1)
```
thp, line 148: Since you only care about whether there's history data or not, a `count_history_data()` or even a `has_history_data()` might speed things up as you mentioned.

lib/urlwatch/command.py
```diff
@@ -142,6 +142,21 @@ def test_filter(self, id):
...
+            if not history_data:
+                logger.info('Add Job: ' + job.pretty_name())
```
thp, line 150:
```suggestion
                logger.info('Add Job: %s', job.pretty_name())
```

lib/urlwatch/command.py
```diff
@@ -142,6 +142,21 @@ def test_filter(self, id):
...
+                new_jobs.append(idx+1)
+
+        if len(new_jobs) != 0:
+            self.urlwatch_config.idx_set = frozenset(int(s) for s in new_jobs)
```
thp, line 154: Why force-convert to `int()` here, wouldn't this also work?

```suggestion
            self.urlwatch_config.idx_set = frozenset(new_jobs)
```

lib/urlwatch/command.py
```diff
@@ -142,6 +142,21 @@ def test_filter(self, id):
         # (ignore_cached) and we do not want to store the newly-retrieved data yet (filter testing)
         return 0
 
+    def run_new_jobs(self):
+        new_jobs = list()
```
thp, line 146:
```suggestion
        new_jobs = []
```

lib/urlwatch/command.py
```diff
@@ -142,6 +142,21 @@ def test_filter(self, id):
...
+        for idx, job in enumerate(self.urlwatcher.jobs):
+            history_data = self.urlwatcher.cache_storage.get_history_data(job.get_guid(), 1)
+            if not history_data:
+                logger.info('Add Job: ' + job.pretty_name())
+                new_jobs.append(idx+1)
+
+        if len(new_jobs) != 0:
```
thp, line 153: This can be simplified as:

```suggestion
        if new_jobs:
```

lib/urlwatch/command.py
```diff
@@ -142,6 +142,21 @@ def test_filter(self, id):
...
+            self.urlwatch_config.idx_set = frozenset(int(s) for s in new_jobs)
+        else:
+            return 0
+        self.urlwatcher.run_jobs()
+        self.urlwatcher.close()
```
thp, line 158: This whole block (with the suggestions from above) can be:

```suggestion
        if not new_jobs:
            return 0
        self.urlwatch_config.idx_set = frozenset(new_jobs)
        self.urlwatcher.run_jobs()
        self.urlwatcher.close()
```

lib/urlwatch/config.py
```diff
@@ -93,6 +93,7 @@ def parse_args(self, cmdline_args):
         group.add_argument('--delete', metavar='JOB', help='delete job by location or index')
         group.add_argument('--enable', metavar='JOB', help='enable job by location or index')
         group.add_argument('--disable', metavar='JOB', help='disable job by location or index')
+        group.add_argument('--run-new-jobs', action='store_true', help='run new jobs or jobs without history')
```
thp, line 96: As you pointed out, it's "new jobs or jobs without history", so maybe it can be something else than `--run-new-jobs`? What about e.g. `--bootstrap-jobs` or `--initialize-jobs` or `--request-first-version` or something (I don't realy like any of them particularly much, but maybe it helps guide you in the right direction).

thp (changes requested):
See comments.

lib/urlwatch/command.py
nille02: Maybe, i did not try it since i haven't looked into minidb. But storage.py:588 is not so long that i think there is much to get. But i can try a bit.

lib/urlwatch/command.py
nille02: I did copy that line from config.py:122. But you are correct its the same result.

lib/urlwatch/config.py
nille02: Yea, i wanted to make it as obviously as it can get.  `--initialize-jobs` is from the three the most fitting in my opinion. 
Or maybe `--prepare-jobs` since its easier to type.

lib/urlwatch/command.py
nille02: I have played a but with it and with the current implementation a look up for 1457 Jobs took around 3 Minutes and 24 Seconds. 

So i did try to tone the get_history_data a bit down but it did not has a huge effect. My next attempt was to broaden the Database queue to not just request the information for a specific guid but everything with tries==0 or None. 
If i do that for each job, the time would be 5 time as much so i did try to queue just once and store the guid's in a list and just reuse it. 
That works great in my first tests and reduced the time to process all these jobs to ~0.875620 seconds.

```
    def has_history_data(self, guid):
        if not self._cached_has_history_data_list:
            for data in list(CacheEntry.query(self.db, CacheEntry.c.guid, where=((CacheEntry.c.tries == 0) | (CacheEntry.c.tries == None)))):  # noqa:E711
                self._cached_has_history_data_list.append(data[0])
        try:
            if self._cached_has_history_data_list.index(guid):
                return True
        except ValueError:
            return False
        return False
```

My main issue now is that i never does anything with redis so far. so a redis version needs a bit more time. 

PS: My current Database has 67432 Rows and its size is 1.026.822.144 Bytes

lib/urlwatch/storage.py
```diff
@@ -600,6 +602,18 @@ def get_history_data(self, guid, count=1):
                     break
         return history
 
+    def has_history_data(self, guid):
+        if not self._cached_has_history_data_list:
+            for data in list(CacheEntry.query(self.db, CacheEntry.c.guid,
+                                              where=((CacheEntry.c.tries == 0) | (CacheEntry.c.tries == None)))):  # noqa:E711
+                self._cached_has_history_data_list.append(data[0])
+        try:
+            if self._cached_has_history_data_list.index(guid):
```
thp, line 611: As you're only using this to query, maybe use a [`set()`](https://docs.python.org/3/library/stdtypes.html#set) and use `.add()` to add and `guid in ...` to check membership.

nille02: Do you mean something like this? 

```python
        if not self._cached_has_history_data_list:
            self._cached_has_history_data_list = frozenset(guid[0] for guid in list(CacheEntry.query(self.db,CacheEntry.c.guid, where=((CacheEntry.c.tries == 0) | (CacheEntry.c.tries == None))))) # noqa:E711
        for entry in self._cached_has_history_data_list:
            if entry == guid:
                return True
            else:
                continue
        return False
```

nille02:
I added the redis version of has_history_data() to it but it just wraps around get_history_data() since redis is already fast. 

For this a new cache.db was used and a redis server that runs on a debian stable vm.
```
Redis          Benchmark: 864 Jobs took 0:00:00.680740
sqllite        Benchmark: 864 Jobs took 0:00:30.835610
sqllite Cached Benchmark: 864 Jobs took 0:00:01.301602
```
The time is just the loop for all jobs over has/get_history_data()

lib/urlwatch/config.py
```diff
@@ -93,6 +93,7 @@ def parse_args(self, cmdline_args):
         group.add_argument('--delete', metavar='JOB', help='delete job by location or index')
         group.add_argument('--enable', metavar='JOB', help='enable job by location or index')
         group.add_argument('--disable', metavar='JOB', help='disable job by location or index')
+        group.add_argument('--prepare-jobs', action='store_true', help='run new jobs or jobs without history')
```
thp, line 96:
```suggestion
        group.add_argument('--prepare-jobs', action='store_true', help='run jobs without history')
```

lib/urlwatch/storage.py
```diff
@@ -688,6 +703,11 @@ def get_history_data(self, guid, count=1):
                         break
         return history
 
+    def has_history_data(self, guid):
+        if self.get_history_data(guid):
+            return True
+        return False
```
thp, line 709:
```suggestion
        return bool(self.get_history_data(guid))
```

lib/urlwatch/storage.py
```diff
@@ -600,6 +602,19 @@ def get_history_data(self, guid, count=1):
...
+            self._cached_has_history_data_set = frozenset(guid[0] for guid in
+                                                           list(CacheEntry.query(self.db,CacheEntry.c.guid,
+                                                                                 where=((CacheEntry.c.tries == 0) |
+                                                                                        (CacheEntry.c.tries == None))))) # noqa:E711
+        for entry in self._cached_has_history_data_set:
+            if entry == guid:
+                return True
+            else:
+                continue
+        return False
```
thp, line 616:
```suggestion
        return guid in self._cached_has_history_data_set
```

lib/urlwatch/storage.py
```diff
@@ -600,6 +602,19 @@ def get_history_data(self, guid, count=1):
                     break
         return history
 
+    def has_history_data(self, guid):
+        if not self._cached_has_history_data_set:
+            self._cached_has_history_data_set = frozenset(guid[0] for guid in
+                                                           list(CacheEntry.query(self.db,CacheEntry.c.guid,
+                                                                                 where=((CacheEntry.c.tries == 0) |
+                                                                                        (CacheEntry.c.tries == None))))) # noqa:E711
```
thp, line 610: Wouldn't we just need to `SELECT DISTINCT guid FROM CacheEntry`? (untested, but probably working...)

```suggestion
        if not self._cached_has_history_data_set:
            self._cached_has_history_data_set = frozenset(CacheEntry.query(self.db, CacheEntry.c.guid.distinct("guid")))
```

thp: Alternatively, drop the whole caching thing and just query the count directly:

```suggestion
        return next(CacheEntry.query(self.db, minidb.func.count("*")("count"), where=CacheEntry.c.guid == guid)).count > 0
```

thp (changes requested):
See comments.

lib/urlwatch/storage.py
nille02: The first works but it still needs to loop over the result since its a "RowProxy" otherwise. 

`self._cached_has_history_data_set = frozenset(guid[0] for guid in list(CacheEntry.query(self.db, CacheEntry.c.guid.distinct("guid"))))  # noqa:E711` seems to work.

The second works but it is just as slow as using get_history_data() direct
`Benchmark: Queue 4059 Jobs took 0:02:07.422035` and cached 
`Benchmark: Queue 4059 Jobs took 0:00:00.194317`

nille02: I did try some other cases with the first but it won't work. 
If a Job failed, i still has its guid in the database with no data and a try count greater than 0

thp (changes requested):
Works for me. Please also add a changelog entry to the PR.

nille02:
Just a entry in the CHANGELOG.md? If yes its done. 

And everyone i hope you all had a happy new year.

thp (approved):
Looks good, thanks for the updates!

thp:
@nille02 Can you fix up the last remaining code style issues?

```
----------------------------- Captured stdout call -----------------------------
/home/runner/work/urlwatch/urlwatch/lib/urlwatch/command.py:151:36: E226 missing whitespace around arithmetic operator
/home/runner/work/urlwatch/urlwatch/lib/urlwatch/storage.py:608:89: E231 missing whitespace after ','
/home/runner/work/urlwatch/urlwatch/lib/urlwatch/storage.py:609:115: W504 line break after binary operator
/home/runner/work/urlwatch/urlwatch/lib/urlwatch/storage.py:610:121: E261 at least two spaces before inline comment
=========================== short test summary info ============================
```

After that, this is ready to merge :)

nille02:
Should be fixed now

thp:
There's still 3 more:

```
/home/runner/work/urlwatch/urlwatch/lib/urlwatch/storage.py:608:89: E231 missing whitespace after ','
/home/runner/work/urlwatch/urlwatch/lib/urlwatch/storage.py:608:108: W291 trailing whitespace
/home/runner/work/urlwatch/urlwatch/lib/urlwatch/storage.py:609:114: W291 trailing whitespace
```

(you can run this locally to verify it's fixed, no need to wait for CI -- use `pytest -v` after installing dependencies, see https://github.com/thp/urlwatch/blob/master/.github/workflows/unit-tests.yml)

nille02:
No it should be done. 

I did try pytest but i got some different error and i guess its due to windows.

thp:
Thanks for working on this and seeing it through! Merged :)

PR #836, opened 2025-01-07
Add option to preserve ids that are provided as arguments with --prepare-jobs

nille02:
If you try to combine `--prepare-jobs` and running Job IDs in one command the ids got ignore since prepare_jobs overrides the set.

With this small change we use a set instead of a list we use a set and merge them later together to preserve other scheduled jobs.

With this we can use for example `urlwatch 1 2 3 4 --prepare-jobs`

### DIFF
--- a/lib/urlwatch/command.py
+++ b/lib/urlwatch/command.py
@@ -143,15 +143,15 @@ class UrlwatchCommand:
         return 0
 
     def prepare_jobs(self):
-        new_jobs = []
+        new_jobs = {*()}
         for idx, job in enumerate(self.urlwatcher.jobs):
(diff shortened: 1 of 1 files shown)


lib/urlwatch/command.py
```diff
@@ -143,15 +143,15 @@ def test_filter(self, id):
         return 0
 
     def prepare_jobs(self):
-        new_jobs = []
+        new_jobs = {*()}
```
thp, line 146: Funky. Just go with `set()` instead, I'd say. Same number of characters, and less confusing:

```suggestion
        new_jobs = set()
```

Issue #835, opened 2025-01-06
Reporters firing even when no jobs run


MarcPer:
I saw the `shell` reporter firing even when there were no changes to the pages I'm monitoring, so I tried disabling all jobs, and it still keeps firing:

```shell
$ urlwatch --verbose

2025-01-06 21:35:41,828 cli INFO: turning on verbose logging mode
2025-01-06 21:35:41,833 minidb DEBUG: PRAGMA table_info(CacheEntry)
2025-01-06 21:35:41,834 main INFO: Using .../urls.yaml as URLs file
2025-01-06 21:35:41,834 main INFO: Using .../hooks.py for hooks
2025-01-06 21:35:41,834 main INFO: Using .../cache.db as cache database
2025-01-06 21:35:41,835 main INFO: Found 2 jobs
2025-01-06 21:35:41,836 worker DEBUG: Processing 0 jobs (out of 2)
2025-01-06 21:35:41,836 reporters INFO: Submitting with stdout (<class 'urlwatch.reporters.StdoutReporter'>)
2025-01-06 21:35:41,836 reporters INFO: Submitting with shell (<class 'urlwatch.reporters.ShellReporter'>)
```

This is the relevant part of my config:
```yaml
display:
  empty-diff: false
  error: true
  new: true
  unchanged: false
job_defaults:
  all: {}
  browser: {}
  shell: {}
  url: {}
report:
  shell:
    command: ['osascript', '-e', 'display notification "Something happened"']
    enabled: true
    ignore_stderr: false
    ignore_stdout: true
```

Is this expected behavior or am I missing something?

MarcPer:
Some additional info:

- urlwatch version: 2.29
- MacOS Version: 15.2 (24C101)
- Python version: 3.13.1
- urlwatch installed with `uv tool install urlwatch`

thp:
"Submitting with" is always printed, but the shell reporter should then not call the script if the diff text is empty:

https://github.com/thp/urlwatch/blob/30f143b1dfa619e34307d49db657828e3d7471ba/lib/urlwatch/reporters.py#L1118-L1120

However, it's weird that it says:

> worker DEBUG: Processing 0 jobs (out of 2)

MarcPer:
> "Submitting with" is always printed, but the shell reporter should then not call the script if the diff text is empty:
> 
[...]

PR #837, opened 2025-02-08
Fix shell reporter running even when diff is empty

MarcPer:
Addresses thp/urlwatch#835


thp (approved):
Well spotted! Thanks a lot :)

Issue #385, opened 2019-04-20
Shorter headings for text report details


lolilolicon:
With this `urls.yaml`:

    name: Firefox for Android
    command: "curl -s -I -o /dev/null -w '%{redirect_url}' 'https://download.mozilla.org/?product=fennec-latest&os=android&lang=multi'"

`urlwatch` produces:

    ===========================================================================
    01. NEW: Firefox for Android
    ===========================================================================

    ---------------------------------------------------------------------------
    NEW: Firefox for Android (curl -s -I -o /dev/null -w '%{redirect_url}' 'https://download.mozilla.org/?product=fennec-latest&os=android&lang=multi')
    ---------------------------------------------------------------------------

You'll agree with me that the second heading is much too long.

There're two aspects to consider. First, the part within the parentheses is
really long, because `jobs.ShellJob.get_location` returns the full `command`.
I think the URL alone is sufficient. We'd then get this:

    ---------------------------------------------------------------------------
    NEW: Firefox for Android (https://download.mozilla.org/?product=fennec-latest&os=android&lang=multi)
    ---------------------------------------------------------------------------

Still too long. The second thing we can do is put the URL on its own line.
We now have this:

    ---------------------------------------------------------------------------
    NEW: Firefox for Android
    https://download.mozilla.org/?product=fennec-latest&os=android&lang=multi
    ---------------------------------------------------------------------------

Much nicer, isn't it?

We don't have to stop there. Did you know about `ls --hyperlink`? If your
terminal supports it, you can right click on the file names and "open link"!
Find out more here: [Hyperlinks (a.k.a. HTML-like anchors) in terminal emulators](https://gist.github.com/egmontkob/eb114294efbcd5adb1944c9f3cb5feda).

For our example, try this in your terminal:

    printf '\e]8;;https://download.mozilla.org/?product=fennec-latest&os=android&lang=multi\e\\Firefox for Android\e]8;;\e\\\n'

So for the `stdout` report, when stdout is a terminal, the heading can be
shortened to this clickable text:

    ---------------------------------------------------------------------------
    NEW: Firefox for Android
    ---------------------------------------------------------------------------

lolilolicon:
A much simpler solution: simply provide an option, which says that if `name` of a job is specified, the `command` or `url` or whatever will not be added to the heading.

thp:
Closing this as feature request. There are some other requests for customizing text output (e.g. #456, #429) that maybe will be tackled in the future.

[...]

PR #842, opened 2025-04-29
Added option to change title format

E1k3:
This PR adds the option to change the report title format as discussed in https://github.com/thp/urlwatch/issues/385.
I added the option to HtmlReporter, TextReporter and MarkdownReporter, because all of these currently use the "pretty (location)" default format in some way.

I tested the text and markdown reporters and they work like discussed. I couldn't test the html reporter, because I couldn't get my mailserver to accept my smtp password for some reason.

For now, every setting despite "pretty" and "location" falls back to the default format.
If you want an error message instead, let me know.

### DIFF
--- a/lib/urlwatch/reporters.py
+++ b/lib/urlwatch/reporters.py
@@ -217,15 +217,22 @@ class HtmlReporter(ReporterBase):
         </head><body>
         """)
 
+        title_format = cfg.get('title', 'default')
+
         for job_state in self.report.get_filtered_job_states(self.job_states):
             job = job_state.job
 
             if job.location_is_url():
                 title = '<a href="{location}">{pretty_name}</a>'
-            elif job.pretty_name() != job.get_location():
-                title = '<span title="{location}">{pretty_name}</span>'
-            else:
+            elif title_format == "location":
                 title = '{location}'
+            elif title_format == "pretty":
+                title = '{pretty_name}'
+            else:  # default
+                if job.pretty_name() != job.get_location():
+                    title = '<span title="{location}">{pretty_name}</span>'
+                else:
+                    title = '{location}'
             title = '<h2><span class="verb">{verb}:</span> ' + title + '</h2>'
 
             yield SafeHtml(title).format(verb=job_state.verb,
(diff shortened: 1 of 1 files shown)


thp (changes requested):
The `cfg.get()` call with a default is great for existing configurations that lack that entry.

New configs should however have `title: default` in there, this makes it also a bit easier to discover this feature (add it to the relevant entries of `DEFAULT_CONFIG` in `storage.py`).

Since the gotify reporter already uses `title`, and we might want to use `title` for something else in the future, AND you already call it `title_format` in the Python variable that gets used (for good reason: it's the _title format_, not the title itself), I'd suggest we call it `title_format` in the config as well.

Also, there should be added documentation for this feature, so that users can find out about how it works and what possible values exist for `title_format` and to which reporters it applies. 

Another thing I would do to detect misconfigurations (instead of accepting them silently):

```python3
if title_format not in ("default", "location", "pretty"):
    logger.warn(f'Invalid title format {title_format!r}, falling back to "default".')
    title_format = "default"
```

Otherwise, if I set `title_format: foo`, it will happily accept this without complaint, and use the default format. If a future version adds e.g. `"fancy"` as a title format, and someone tries to use it, but uses a version that doesn't have that title format supported yet, instead of silently falling back to `"default"`, it will still fall back to `"default"` (a sensible behaviour), but also write out a warning, giving a clue to the user that the program has done the fallback, because it doesn't recognize the value set.

E1k3:
I added the default options and the warning.
Regarding the documentation: As far as I can see, there is currently no place where the Text, Html and Markdown reporters are individually documented.
The Markdown reporter is somewhat documented in the [Matrix reporter section](https://github.com/thp/urlwatch/blob/d57ae4ea3d2a70562ffa532bfa3d3c94c276f2b2/docs/source/reporters.rst?plain=1#L314) and the Html and Text reporter options are shown without explanations in the [Reporters section of the Configuration page](https://github.com/thp/urlwatch/blob/d57ae4ea3d2a70562ffa532bfa3d3c94c276f2b2/docs/source/configuration.rst?plain=1#L102).
I can add the title_format documentation to these locations or create new sections on the Reporters page for all three reporters.
What do you prefer?

thp:
> @thp I added the format options to the places I mentioned. If you would like more elaborate documentation of the title_format feature, let me know where in the docs that would make the most sense.

Probably a section in the https://github.com/thp/urlwatch/blob/master/docs/source/advanced.rst page, something like "Customizing the title format in reports", with all available options and examples what they do.

thp:
Closing stale PR.

PR #843, opened 2025-05-06
Fix minor bugs on tags filter

mwerlen:
Hi,

A debian user opened [a bug](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1104785) concerning the tags filter.
It seems there is no protection when some jobs don't have tags, `self.tags` is then `None` and it throws a `TypeError`.

This PR is a fix for this specific case.
I also fixed a small error when a tag was set as a string instead of an array.

Regards

Maxime

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ The format mostly follows [Keep a Changelog](http://keepachangelog.com/en/1.0.0/
 
 - Fix shell reporter running even when diff is empty (#837, by MarcPer)
 - Filter for gitlab.com tags fixed (#839, by julianuu)
+- Fix `TypeError` when jobs don't have tags (#843 by Maxime Werlen)
 
 ## [2.29] -- 2024-10-28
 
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -50,6 +50,7 @@ The format mostly follows [Keep a Changelog](http://keepachangelog.com/en/1.0.0/
 - Fix HTML diff table rendering for long line lengths (#793 by trevorshannon)
 - Fix IndexError after failed edit (#801 by jwilk)
 - Fix concurrency issue in Python 3.12 by upgrading to minidb 2.0.8 (fixes #779)
+- Fix some minors bugs on the tags filter (#843 by Maxime Werlen)
```
thp, line 53: This should go into the `[UNRELEASED]` section, it seems to have slipped into the `[2.29]` section by accident :)

Also, maybe:


```suggestion
- Fix `TypeError` when jobs don't have tags (#843 by Maxime Werlen)
```

thp (approved):
Thanks for upstreaming this patch! Just the CHANGELOG.md entry needs to be moved up (and possibly edited), everything else looks great as-is.

CHANGELOG.md
mwerlen: Thanks for the review.
I've moved the entry to the UNRELEASED section.

PR #858, opened 2025-12-10
Prevent Markdown reporter crash when max_length is None

PeterDaveHello:
Guard the None max_length case so markdown reporter no longer raises TypeError on enablement. Add reporter tests for unlimited and trimmed paths to prevent regression.

GitHub Copilot summary:

> This pull request introduces a minor fix to the `MarkdownReporter.submit` method and adds new unit tests to ensure correct handling of the `max_length` parameter. The main focus is to prevent errors when `max_length` is `None` and to verify that report trimming works as expected.
> 
> Testing improvements:
> 
> * Added a new test file `lib/urlwatch/tests/test_reporters.py` with unit tests for `MarkdownReporter.submit`, including scenarios where `max_length` is `None` and where output trimming occurs.
> 
> Bug fix:
> 
> * Updated `MarkdownReporter.submit` in `lib/urlwatch/reporters.py` to only subtract the trimmed message length from `max_length` if `max_length` is not `None`, preventing possible errors when `max_length` is unspecified.

### DIFF
--- a/lib/urlwatch/reporters.py
+++ b/lib/urlwatch/reporters.py
@@ -824,11 +824,24 @@ class MarkdownReporter(ReporterBase):
             details = None
 
         trimmed_msg = "*Parts of the report were omitted due to message length.*\n"
-        max_length -= len(trimmed_msg)
+        summary_raw = summary
+        details_raw = details
+        footer_raw = footer
 
         trimmed, summary, details, footer = MarkdownReporter._render(
-            max_length, summary, details, footer
+            max_length, summary_raw, details_raw, footer_raw
         )
+
+        trimmed_marker = trimmed_msg
+        if max_length is not None and trimmed:
+            if len(trimmed_marker) > max_length:
+                trimmed_marker = "[...]\n"
+                if len(trimmed_marker) > max_length:
+                    trimmed_marker = ""
+            reserved_length = max_length - len(trimmed_marker) if trimmed_marker else max_length
+            trimmed, summary, details, footer = MarkdownReporter._render(
+                reserved_length, summary_raw, details_raw, footer_raw
+            )
 
         if summary:
             yield from summary
@@ -840,8 +853,8 @@ class MarkdownReporter(ReporterBase):
                 yield body
                 yield ''
 
-        if trimmed:
-            yield trimmed_msg
+        if trimmed and trimmed_marker:
+            yield trimmed_marker
 
         if summary and show_footer:
             yield from footer
--- a/lib/urlwatch/tests/test_reporters.py
+++ b/lib/urlwatch/tests/test_reporters.py
@@ -1,0 +1,124 @@
+import datetime
+
+from urlwatch.reporters import MarkdownReporter
+
+
+class _DummyJob:
+    def pretty_name(self):
+        return "dummy"
+
+    def get_location(self):
+        return "dummy"
+
+
+class _DummyJobState:
+    def __init__(self, verb="changed", old_data="old", new_data="new", diff="diff"):
+        self.job = _DummyJob()
+        self.verb = verb
+        self.old_data = old_data
+        self.new_data = new_data
+        self._diff = diff
+        self.traceback = "traceback"
+
+    def get_diff(self):
+        return self._diff
+
+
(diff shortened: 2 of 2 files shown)


thp (changes requested):
In which situation would `max_length` actually become `None`?

lib/urlwatch/tests/test_reporters.py
```diff
@@ -0,0 +1,60 @@
...
+
+    assert output == []
+
+
+def test_markdown_submit_with_trimming_returns_notice():
+    report = _DummyReport()
+    job_state = _DummyJobState(diff="line1\nline2\nline3")
+    reporter = MarkdownReporter(report, report.config["report"]["markdown"], [job_state], datetime.timedelta())
+
+    output = list(reporter.submit(max_length=10))
```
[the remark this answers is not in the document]
PeterDaveHello: It only goes negative if a very small max_length is passed. This is because submit() first does `max_length -= len(trimmed_msg)` and the trimmed_msg string is about 58 chars, so max_length=10 becomes -48. Even then it just trims and does not crash. If you want to avoid that impression, I can change the test to use a value >= 60.

PeterDaveHello:
Agreed that a plain CLI run won't hit this. The crash only happens when the Markdown reporter output is actually consumed (Python/hook/extension usage): with default `max_length=None`, `submit()` subtracts from None. This is a minimal guard and does not change behavior when a length is provided.

thp (changes requested):
The more I think about it, the more I think that `max_length -= len(trimmed_msg)` should be done later (only when `trimmed` is truthy). If it isn't trimmed, we still want to have the full `max_length`, and if it is trimmed, something like `[...]` or `[trimmed]` might be better to allow for more "payload" and less "boilerplate". In particular, if `max_length < len(trimmed_msg)`, then technically we wouldn't be able to show anything, not even the summary or the details. And if the "trimmed message" is quite small (e.g. "[...]" - 5 characters).

Then again, it's probably the `_render()` method that should do all deciding and value adding (possibly with an additional return value before the footer), so that it knows what the "real" max length is, and either fit the message in there, or trim it.

In other words:

The message "Hello world" (11 characters) and `trimmed_msg = "[...]"` (5 characters)

with max_length == 11 should still be:

"Hello world"

but with max_length == 10, should be:

"Hello[...]"

Right now, with max_length == 11, it would be:

"Hello [...]"

lib/urlwatch/reporters.py
```diff
@@ -824,12 +824,25 @@ def submit(self, max_length=None):
...
+            max_length, summary_raw, details_raw, footer_raw
         )
 
+        trimmed_marker = trimmed_msg
+        if max_length is not None and trimmed:
+            if len(trimmed_marker) > max_length:
+                trimmed_marker = "[...]\n"
+                if len(trimmed_marker) > max_length:
+                    trimmed_marker = ""
+            reserved_length = max_length - len(trimmed_marker) if trimmed_marker else max_length
```
thp, line 841: If `trimmed_marker` is the empty string (the only way `if trimmed_marker` should evaluate to falsy), then `len(trimmed_marker) == 0`, so the result is the same.

```suggestion
            reserved_length = max_length - len(trimmed_marker)
```

lib/urlwatch/reporters.py
```diff
@@ -824,12 +824,25 @@ def submit(self, max_length=None):
...
+            trimmed, summary, details, footer = MarkdownReporter._render(
```
thp, line 842: It's unfortunate that we have to call `MarkdownReporter._render()` twice here. Can't `MarkdownReporter._render()` contain all that trimming logic?

thp (changes requested):
See comments. Ideally call `MarkdownReporter._render()` only once and let it figure out trimming (since `MarkdownReporter._render()` already returns the `trimmed` status, might as well return `trimmed_msg` instead (being the empty string if not trimmed or not enough space).

thp:
@PeterDaveHello Any updates here?

PeterDaveHello:
Sorry, I forgot this PR, will take a look at it.

PR #861, opened 2026-01-25
Add browser_path option for browser jobs

Matt-Deacalion:
Allow specifying a browser executable for Playwright instead of the bundled browser.

The option can be set per job or globally via job_defaults in urlwatch.yaml, and supports system-installed browsers or custom builds.


thp (changes requested):
Please revert the manpage changes and the mode change on `update-manpages.sh`. The manpages are usually updated in one go during release.

As for the name of the key, what about `browser_executable`, or `browser_executable_path`? It more closely matches what playwright takes as a parameter, and doesn't confuse with the concept of "path" as in e.g. search path (`$PATH` and otherwise).

I'm not ure about the `urlwatch.yaml` example and the addition to `urls.yaml`, it feels like this would be more discoverable if there was an advanced topic in `docs/source/advanced.rst` that describes how to fully use that feature (including per job or globally) and which browser (or browser families) are supported. E.g. does it work with Firefox and derivatives? Chrome and derivatives? Any other browsers? Distro-installed vs Flatpak vs ... (probably fine to just link to the Playwright documentation for this, if there is any.

PR #866, opened 2026-05-06
Add json option for url jobs

sbraz:
Hi,

When set alongside data, serializes the payload as JSON and sets
Content-type: application/json. This lets users specify JSON request
bodies as native YAML mappings instead of embedding escaped JSON
strings in data.

### DIFF
--- a/lib/urlwatch/jobs.py
+++ b/lib/urlwatch/jobs.py
@@ -280,9 +280,9 @@ class UrlJob(Job):
     __kind__ = 'url'
 
     __required__ = ('url',)
-    __optional__ = ('cookies', 'data', 'method', 'ssl_no_verify', 'ignore_cached', 'http_proxy', 'https_proxy',
-                    'headers', 'ignore_connection_errors', 'ignore_http_error_codes', 'encoding', 'timeout',
-                    'ignore_timeout_errors', 'ignore_too_many_redirects', 'ignore_incomplete_reads')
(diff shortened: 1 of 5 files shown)


lib/urlwatch/jobs.py
```diff
@@ -314,12 +314,21 @@ def retrieve(self, job_state):
             headers['Cache-Control'] = 'max-age=172800'
             headers['Expires'] = email.utils.formatdate()
 
+        if self.data is not None and self.json is not None:
+            raise ValueError('Cannot specify both "data" and "json" for the same job')
```
thp: One way to [prevent this error](https://en.wikipedia.org/wiki/Poka-yoke) would be to allow data always to be specified via the `data` key, but just either:

 - Introduce a new boolean (default `false`) `use_json` (or just `json`?) field, if it's `true`, will set content type
 - Introduce a new string `content-type` (default `application/x-www-form-urlencoded`) that can be set to `application/json` (but also other values, for maximum flexibility)

In both cases, the data would be specified via the `data` key.

sbraz: Good idea, I'm leaning towards `json: true`, it's more explicit than relying on the `content-type`. I'll update the PR accordingly.

sbraz: Pushed and updated the description, let me know if you're OK with this implementation.

thp (approved):
Nice and clean, thanks!

PR #868, opened 2026-06-25
Reports a browser job error for an HTTP error response

### LINKED ISSUE #867: `navigate` jobs do not report errors like `url` jobs do.
When urlwatch gets an HTTP error response (such as 404) from a server, that job will be reported as an error if the job is a `url` type job.  If it is a `navigate` type job, the error will not surface.  Page content is returned no matter the HTTP status code.

trevorshannon:
Introduces a browser job option `raise_for_status` that, when true, will raise an error if the response has an HTTP error status.  Omitting the option results in the same behavior as before this change.

Once the browser navigates to the page of interest (following any redirects), if the resulting HTTP status code is not in the range 200-299, an error is raised.   This error is reported the same as an HTTP error in a `url` job: no traceback, showing only the status code & message.

This is particularly useful for keeping temporary `404` or `500` errors from polluting the job history for browser jobs.

fixes #867


lib/urlwatch/jobs.py
```diff
@@ -455,10 +461,19 @@ def retrieve(self, job_state):
...
                 locator = page.locator(self.wait_for)
                 locator.wait_for()
 
             return page.content()
+
+    def format_error(self, exception, tb):
+        if isinstance(exception, BrowserHttpError):
+            # A full traceback is not necessary for this type of error.
+            return str(exception)
+        return tb
```
thp: While we do this above (in `UrlJob.format_error()`, the return at the end (in `UrlJob` and new here in `BrowserJob`) should probably call `super().format_error(exception, eb)` (which is functionally equivalent to `return tb`, since that's what `JobBase.format_error()` does), but conceptually clearer.

docs/source/jobs.rst
```diff
@@ -97,6 +97,9 @@ Job-specific optional keys:
 - ``useragent``: ``User-Agent`` header used for requests (otherwise browser default is used)
 - ``browser``:  Either ``chromium``, ``chrome``, ``chrome-beta``, ``msedge``,
   ``msedge-beta``, ``msedge-dev``, ``firefox``, ``webkit`` (must be installed with ``playwright install``)
+- ``raise_for_status``: If true, reports an error if the HTTP response of the
+  final page after redirection is not OK, i.e. with a status code outside of
+  200-299.
```
thp: For backwards compatibility with existing installations, it's probably a good idea to make it opt-in.

However, in the future, it might be more consistent to use `ignore_http_error_codes` similar to `UrlJob`, and by default raise an error.

The two "problematic" situations I see:

Case A
------

 * A user having used urlwatch before with browser jobs has previously successfully watched a page that e.g. 404s, and now (with raise-by-default) gets an error. They would need to add 404 to `ignore_http_error_codes` (just like they would need to do if they were using `UrlJob`)

Case B
------

* A new user uses urlwatch and doesn't do any special `raise_for_status` configuration. HTTP errors are silently ignored and treated as "empty page" (or page changes, in any case - they don't raise exceptions)

In fact, https://github.com/thp/urlwatch/issues/867 is basically case B, not case A.

For consistency and future-proofing (case A is basically a one-off migration issue, case B is something that lingers for forever, and is more likely to happen as time marches one and [uncool URIs change](https://www.w3.org/Provider/Style/URI)) I think going with error-by-default is a sane option.

So I'd be happy with the backwards incompatible change (it can be documented in `docs/source/deprecated.rst`, but probably doesn't need to be) and making it so that the behaviour you seek is the new default, and opting out of the behavior is done via `ignore_http_error_codes` (and `ignore_error()` override?). This might not be 100% what the previous behavior is (treating error pages as normal pages), but if that need comes up, we can always patch it in later).

thp (changes requested):
See comments. I think it's worthwhile to make the desired behaviour the default, as the option is otherwise a "fix the (buggy) legacy behavior" toggle that new users need enable to fix behaviour.

For consistency, having `ignore_http_error_codes` like `UrlJob` seems like a good idea.

We can always add [spacebar heating](https://xkcd.com/1172/) back as an option later if someone complains (or not..).

lib/urlwatch/jobs.py
trevorshannon: good idea, done

docs/source/jobs.rst
trevorshannon: Ok great, I like the errors by default better.

Case B is indeed what I was thinking about.  Case A seems pretty easy to deal with as you say.  Do let me know if you hear about people needing to watch the content of an e.g. 404 not found page.

I opted to not update `deprecated.rst`

lib/urlwatch/tests/test_handler.py
```diff
@@ -331,3 +333,30 @@ def test_reset_tries_to_zero_when_successful():
...
             cache_storage.close()
+
+
+@pytest.mark.parametrize('status_code, ignore_http_error_codes, expect_error, expect_ignored', [
+    (404, '4xx', True, True),
+    (404, None, True, False),
+    (200, '4xx', False, False),
+    (200, None, False, False),
+])
+def test_browser_job_ignores_http_error(
```
trevorshannon, line 344: Most of the other tests in this file are a bit more integration-y than I felt was needed for this test.  If you'd like me to follow the outline of the existing tests & actually call `urlwatcher.run_jobs()` with a YAML file & everything, I can do that instead.

lib/urlwatch/util.py
```diff
@@ -138,3 +138,21 @@ def chunkstring(string, length, *, numbering=False):
         return ('{} ({}/{})'.format(part, i + 1, len(parts)) for i, part in enumerate(parts))
 
     return (string[i:length + i].strip() for i in range(0, len(string), length))
+
+
+def ignore_http_code(status_code, codes_to_ignore):
```
thp: The verb of the function name is really "should ignore", not "ignore":

```suggestion
def should_ignore_http_code(status_code, codes_to_ignore) -> bool:
```

lib/urlwatch/util.py
```diff
@@ -138,3 +138,21 @@ def chunkstring(string, length, *, numbering=False):
...
+      '4xx' are accepted.
+    """
+    ignored_codes = []
+    if isinstance(codes_to_ignore, int) and codes_to_ignore == status_code:
+        return True
+    elif isinstance(codes_to_ignore, str):
+        ignored_codes = [s.strip().lower() for s in codes_to_ignore.split(',')]
+    elif isinstance(codes_to_ignore, list):
+        ignored_codes = [str(s).strip().lower() for s in codes_to_ignore]
+    return str(status_code) in ignored_codes or '%sxx' % (status_code // 100) in ignored_codes
```
thp: Let's use modern formatting here (it took me a double-take to not interpret %sxx as some kind of hexadecimal formatting, but a "%s" followed by a literal "xx", even though it was like that before already):

```suggestion
    return str(status_code) in ignored_codes or f'{status_code // 100}xx' in ignored_codes
```

lib/urlwatch/jobs.py
```diff
@@ -453,10 +454,24 @@ def retrieve(self, job_state):
...
 
             return page.content()
+
+    def format_error(self, exception, tb):
+        if isinstance(exception, BrowserHttpError):
+            # A full traceback is not necessary for this type of error.
+            return str(exception)
+        super().format_error(exception, tb)
+
+    def ignore_error(self, exception):
```
thp: Similar to below, the verb isn't "ignore", but "should ignore":


```suggestion
    def should_ignore_error(self, exception) -> bool:
```

In other words, the function doesn't do any ignoring by itself, but rather answers the question whether or not the error should be ignored by the caller of the function.

docs/source/advanced.rst
```diff
@@ -167,18 +167,18 @@ the ``ignore_connection_errors`` key in the job list configuration file:
    ignore_connection_errors: true
 
 Similarly, you might want to ignore some (temporary) HTTP errors on the
-server side:
+server side. This can be done for ``url`` or ``navigate`` jobs:
```
thp: The job kinds are called "url" and "browser", the key is `url` and `navigate` (unfortunately). I'd say this line here should be:


```suggestion
server side. This can be done for ``url`` or ``browser`` jobs:
```

And then below, instead of saying "navigate job also ok here", say "key can be 'navigate' for browser jobs".

thp (changes requested):
Just minor style and wording comments to make it super clean. Rest looks good now. Thanks for working on this!

lib/urlwatch/util.py
trevorshannon: haha me too, that took me a while to parse out when I was reading through the `UrlJob` code.  Your suggestion helps.

thp (approved):
Looks good to me now. Thanks!

Issue #869, opened 2026-06-29
fingerprint-real firefox option for the browser job?


feder-cr:
the browser job already renders pages with playwright, which is great for js-heavy sites, but pages that filter automated browsers get monitored as a challenge or empty page rather than the real content. would an option to point the browser job at a fingerprint-real firefox build help there? something like a custom executable/launcher so users can plug in invisible_playwright (a drop-in playwright wrapper over a patched firefox with a realistic, consistent fingerprint), default unchanged. happy to send a PR if it's of interest.

thp:
Yes, having that as an option is great. Feel free to create a PR.

PR #871, opened 2026-07-11
Add executable_path to browser jobs

feder-cr:
Adds an optional executable_path to browser jobs, so Playwright can launch a custom browser build (for example a fingerprint-real Firefox) instead of the bundled one. browser already selects the engine; this just adds the binary path and defaults to unset, so current behavior is unchanged. Docs updated. Follows up on #869. Browser job tests pass locally.


thp:
Looks good on the surface, feel free to mark as non-draft when you think it's ready. Also add a `CHANGELOG.md` entry.

thp:
Seems like this is similar to #861, however I like the `executable_path` naming better. #861 has some other yet unresolved review comments, so IMHO if we can fix this one (#871) up, I'd prefer merging #871. cc @Matt-Deacalion

